PR #56, opened 2015-01-24
Attachment Feature

janschaefer:
It is often desirable to be able to add attachments to steps. One common use case are selenium tests where having a screenshot of the current page often helps in debugging a problem.
In addition, it might even be useful for business owners to actually be able to see the content of a page in a scenario. Thus JGiven should provide the possibility to add attachments to steps.
In this pull request this is implemented. Currently attachments are only supported for the HTML5 report.


janschaefer:
Scenarios: http://jgiven.org/jgiven-report/html5/#/tag/Feature/Attachments

PR #68, opened 2015-03-22
reference correct class in code example

ahus1:
Hello Jan,

I found this trying the example. Too bad one can't reference source code from markdown to avoid this (only Asciidoctor AFAIK). 

Best regards,

Alexander


janschaefer:
Thanks. Yes it would be much better to directly include working and tested code in the documentation.

PR #69, opened 2015-04-13
Add cssClass to @IsTag to customize tags in HTML reports

janschaefer:
Add a new attribute `cssClass` to the `@IsTag` annotation.

```
public @interface IsTag {
    String cssClass() default ""; // Or the default can come from Foundation
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # v0.7.2
-* Added `cssClass` to `@IsTag` to customize tags in HTML reports
+
+## New Features
+
+* Added `cssClass` and `color` attributes to `@IsTag` to customize tags in HTML reports [#69](https://github.com/TNG/JGiven/pull/69)
+
+## Fixed Issues
+
+* Custom CSS files are now copied to the target folder when generating HTML5 reports [#70](https://github.com/TNG/JGiven/issues/70)
 
 # v0.7.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+# v0.7.2
+* Added `cssClass` to `@IsTag` to customize tags in HTML reports
+
 # v0.7.1
 
 ## New Features
--- a/jgiven-core/src/main/java/com/tngtech/jgiven/annotation/IsTag.java
+++ b/jgiven-core/src/main/java/com/tngtech/jgiven/annotation/IsTag.java
@@ -83,7 +83,7 @@ public @interface IsTag {
     /**
      * Sets a CSS class that should be used in HTML reports for this tag.
      * <p>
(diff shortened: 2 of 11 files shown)


nobeh:
Thanks for picking this up so quickly! :thumbsup:

janschaefer:
@nobeh if you want to try out the new feature you can do so by using the snapshot repository https://oss.sonatype.org/content/repositories/snapshots/ with JGiven version `0.7.2-SNAPSHOT`

nobeh:
I'm having this tag:

```java
@IsTag(value = "Story", description = "Ticket is a tag to label test scenarios.",
    descriptionGenerator = TicketDescriptionGenerator.class, 
    prependType = false, type = "Story", color = "blue")
@Retention(RetentionPolicy.RUNTIME)
public @interface Story {

  /**
   * @return
   */
  String value();

}
```

and the tag is used like `@Story("CSP-3031")` which is similar to JGiven's ticket URL generator.

then using `html` report, gives me snippets like:

```html
<div class="tag-line">
   <div class="tag tag-Story">
      <a href="Story-CSP-3031.html">CSP-3031</a>
   </div>
</div>
```

and using `html5` report, gives me snippets like:

```html
<span ng-repeat="tag in scenario.tags" class="tag ng-scope">
   <a href="#tag/Story%2FCSP-3031">
      <span class="radius label ng-binding">CSP-3031</span>
   </a>
</span>
```

What am I missing?

janschaefer:
have you updated the version of the jgiven-html5-report dependency to 0.7.2-SNAPSHOT?

nobeh:
Yes, all dependencies are now using 0.7.2-SNAPSHOT, and only JGiven Maven plugin is at 0.7.1.

janschaefer:
I guess you have to update the JGiven maven plugin also to 0.7.2-SNAPSHOT

nobeh:
Trying that yields to:

```
.m2/repository/com/tngtech/jgiven/jgiven-maven-plugin/0.7.2-SNAPSHOT/jgiven-maven-plugin-0.7.2-SNAPSHOT.jar), Plugin's descriptor contains the wrong version: 0.7.1 -> [Help 1]
```

janschaefer:
ok, I will have a look at that and tell you when you can test again

janschaefer:
ok, I have republished the artifacts to the snapshot repository. It should work now.

nobeh:
Confirm that it works as expected. Thanks!
jgiven-tag-colors-0 7 2-snapshot

janschaefer:
Great! Thanks for testing!

nobeh:
Any chance of releasing 0.7.2 soon? :grinning:

janschaefer:
Well, it wouldn't be a very big release, but it wouldn't hurt either ;-)

PR #93, opened 2015-07-21
normalizing encoding and line endings in tests

ahus1:
... to make the build work on Windows. The only other change I needed to do locally was to change `bower` to `bower.cmd` in `jgiven-html5-report\build.gradle`. Don't know yet how to make that flexible.


janschaefer:
Thanks!

PR #94, opened 2015-07-30
Enhanced spring support. 

TripleNail:
Spring beans that are used as stages/step can now be fully managed by spring.

I also added a new JGivenStage annotation so that writing spring based JGiven stages is very simple. Just setup spring for JGiven usage as shown in the examples and annotate your stages with JGivenStage


janschaefer:
Very cool! Thanks a lot for that. The Spring support was very rudimentary so far ;-)

CrystalMethod:
:+1:

jgiven-spring/src/main/java/com/tngtech/jgiven/integration/spring/JGivenStage.java
```diff
@@ -0,0 +1,29 @@
...
+/**
+ * Annotation that marks a bean as JGiven Stage.
+ *
+ * @since 0.8.0
+ */
+@Target({ElementType.TYPE})
+@Retention(RetentionPolicy.RUNTIME)
+@Documented
+@Component
+public @interface JGivenStage {
```
TripleNail: `@JGivenStage` is meant as replacment for `@Component` which also does not use `@Inherited`

TripleNail:
> Does this mean that users have to configure this themselves?

Currently yes, but maybe I will add some kind of auto-configuration facility like `@EnableJGivenSupport`   (and equivalent xml-based method as well) in 0.8.1

jgiven-spring/src/main/java/com/tngtech/jgiven/integration/spring/JGivenStage.java
janschaefer: Ok.

PR #106, opened 2015-08-26
Upgrade to Paranamer 2.8

janschaefer:
the upgrade fixes an incompatibility with Java 8 that required that for Java 8 JGiven had to use reflection based on the -parameters option of the javac compiler. With the new Paranamer version 2.8 this is not necessary anymore.


janschaefer:
See https://github.com/paul-hammant/paranamer/issues/17 for details

PR #153, opened 2015-10-26
spring autoconfiguration for JGiven

TripleNail:
As suggested by Jan there should be less boilerplate code needed for JGiven users to enable spring support. Use `@EnableJGiven` in case of a java based spring configuration, or use `<jgiven:annotation-driven />` in case of an xml based spring configuration.

See unit tests of jgiven-spring package for more examples.


janschaefer:
Very cool, thanks a lot!

PR #159, opened 2015-11-02
Tags in html report can be used as direct links to external websites

davidtron:
I would like to be able to annotate my tests with the class of related tests.

My use case is I have a large number of tests that constitute a larger flow and would like to annotate the code with previous or next and use the annotation to create a link that takes me directly to that test in the html report.

As an enhancement I've added href and hrefGenerator to the tag system, and have modified
the HTML report to link directly to that href if present.

```java
@LinkToTest(LinkExampleTwoTest.class)
@Test
public void link_to_another_test() {
    given().test_linked_to_another_test();
    when().the_report_is_generated();
    then().the_link_appears_in_the_report();
}
```

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,1 +1,1 @@
-version=0.9.4
+version=0.9.5-SNAPSHOT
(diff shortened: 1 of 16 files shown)


janschaefer:
I totally agree that this is useful. Thank you for contributing!

PR #165, opened 2015-11-10
support for camel case in steps (fixes #164)

### LINKED ISSUE #164: allow camel case as well as underscore for step methods
Scenarios allow camel case in the name is it possible to allow the same for step methods?

albertofaci:
I followed the same approach as for scenario description. It assumes snake_case if underscore is found. I had to play with the camelCaseToReadableText because we want it capitalised for scenario descriptions but not for steps. :)


janschaefer:
Great, thanks!

Issue #17, opened 2014-09-06
Nested Steps


janschaefer:
The idea of steps is the following. Often you want to hide certain details of a step, because they are not directly important for a certain scenario. For example:
```
when().I_fill_out_the_registration_form_with_valid_values();
```
will fill out the registration form with valid values, but the concrete values are not shown. If a developer is interested in the concrete values she can look into the implementation of the step. However, in the report these details are completely gone. Thus, if someone reads the report and he wants to know the values, he has to ask a developer.

# Nested Steps
Nested steps are steps that happen within another step. For example, the above step could be defined as follows:
```
public SELF I_fill_out_the_registration_form_with_valid_values() {
    return I_enter_as_name("Franky")
        .and().I_enter_as_email_address("franky@acme.com")
        .and().I_enter_as_password("password1234")
        .and().I_enter_as_repeated_password("password1234");
}
```
The idea is now that in the generated report these sub-steps can be made visible somehow. For example by expanding the parent step. That way, the original scenario stays small, but the details can still be looked up.

# Annotation
Nested steps should get an additional annotation `@NestedSteps` to indicate that the step contains sub-steps.

janschaefer:
Thank you for starting this long-wished feature :-). On a first glance it looks pretty good. I will give you more feedback as soon as possible.

PR #174, opened 2015-12-07
implemented nested steps #17 (+ text report only)

albertofaci:
Hello, this is an attempt to support nested steps.

I had to change a few things around ScenarioModelBuilder to keep a stack of steps so we can add a nested step to the right parent and supporting multilevel nesting. 

As for the report, I have just done it for the plain text report. Note that the nested steps are for report purposes only, i.e. the step status, duration, attachments, etc, are really kept at the top level only. Otherwise it gets too complicated.
If this is accepted, a separate issue should be created for html5 report maybe with a collapsible widget

### DIFF
--- a/jgiven-core/src/main/java/com/tngtech/jgiven/annotation/NestedSteps.java
+++ b/jgiven-core/src/main/java/com/tngtech/jgiven/annotation/NestedSteps.java
@@ -1,0 +1,9 @@
+package com.tngtech.jgiven.annotation;
+
+import java.lang.annotation.*;
+
+@Documented
+@Retention( RetentionPolicy.RUNTIME )
+@Target( { ElementType.METHOD } )
+public @interface NestedSteps {
+}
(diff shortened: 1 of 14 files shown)


albertofaci:
Hiya, any comments or suggestions about this?

janschaefer:
Hi, sorry, I already worked a bit on it, but did not came very far yet due to lack of time. You can have a look at the new branch `nested-steps`. I changed a bit the appearence of the console report and I also would like to integrate the correct execution status on the sub steps. I also would like to integrate it into the html report

janschaefer:
I slightly adapted your implementation and also took the chance to refactor the method intercepting code. I also changed a bit the way the nested steps are printed. In particular, I believe the fail status should be printed after the nested step that actually failed.

albertofaci:
Nice one, thank you, Jan!

As far as I can see the failed status will appear only on the parent top most state.

Current
```
Given something with nested steps that fails (failed)
 |  Given something
 |  And another thing
 |  And something else that fails
When something happens (skipped)
Then something has happen (skipped)
something else not (skipped)
```

whereas the following would be desirable?

```
Given something with nested steps that fails (failed)
 |  Given something (passed)
 |  And another thing (failed)
 |  And something else that fails (skipped)
When something happens (skipped)
Then something has happen (skipped)
something else not (skipped)
```

janschaefer:
With the current implementation the parent step will not be shown as failed. But I agree that this makes sense. I will change that.

janschaefer:
I did a first HTML report implementation. Collapsing is still missing though. What do you think? 

nested-steps

albertofaci:
well done, that looks brilliant!

PR #179, opened 2015-12-27
Reorganized dependencies adding Gradle

tedyoung:
Reorganized around the test runner used, adding in Gradle dependency specifications.


janschaefer:
Thanks!

Issue #186, opened 2016-02-04
Reduce the amount of data required for the HTML report


janschaefer:
When loading a HTML report with 2500 Scenarios the transfered data is about 20 MB. This results in long initial loading times. It would be great if the amount could be reduced somehow.

PR #196, opened 2016-03-15
ensure that contents are encoded utf-8 for pako lib

ahus1:
This occurs when running on an OS without UTF-8 as default charset (i.e. windows machines). This problem has been introduced in #186. 
I found it hard to provide an automated test for this. Therefore I looked at JGiven's own report before and after regarding the German test cases. 
Please double-check if this still works on UTF-8 (well, it should, shouldn't it?)


janschaefer:
Thanks!

PR #197, opened 2016-03-15
make test independent of system's CR/CRLF line ending settings

ahus1:
otherwise this test will not run on the windows platform.


janschaefer:
Thanks!

Issue #202, opened 2016-04-18
Stage composition (stages within stages)


albertofaci:
A bit of background: At the moment we have ~500 acceptance tests in JGiven. These are microservice-level component tests that talk over HTTP with REST endpoints. They use a application running with a DB and some mocks abstract away the other microservices. The test setup is thus quite complex involving setting data up that is not specifically relevant for the test but just _needs to be there_. So far we've used inheritance and a lot of stages in the test with `@ScenarioStage`.

We now feel that we need to compose stages within stages. To have stage methods that are a gateway to a few other methods that live in various stages. And we don't need the sub-stages to appear in the report. We have used inheritance in some instances, but it only really allows to decorate the access to one stage. (Note that these are not nested stages, as they live in a different class).

Product owners are a bit picky about cluttering the report with irrelevant steps and the `@Hidden` functionality feels to much of a workaround for this, and we don't want the stage to be hidden always either.

We basically would like to achieve something like this:
```

public class DrivingTest extends SimpleScenarioTest<DrivingStage> {
  
  @ScenarioStage
  CarStage carStage;

  @Test
  public a_car_can_drive() {
       carStage.given().a_car_exists();
       when().the_user_drives_to("London")
       then().he_gets_to("London")
  }
 
}

public class DrivingStage extends Stage<DrivingStage> {
    
    public DrivingStage the_user_drives_to(@Quoted String location) {
    }

    public DrivingStage he_get_to(@Quoted String location) {
    }
}


public class WheelStage extends Stage<WheelStage>  {
   
   @ProvidedScenarioState
    List<Wheel> wheels; //I want to use this somewhere else
   
   public WheelStage $_wheels_exists(int num) {
       wheels = [something]
   }
}

public class ChassisStage extends Stage<ChassisStage>  {
   public ChassisStage a_chassis_exists() {
   ...
   }
}

//***** here comes the magic
public class CarStage extends Stage<CarStage> {
  
  @ScenarioStage //or @SubStage or some other annotation
  WheelStage wheelStage;

  @ScenarioStage
   ChassisStage chassisStage;

  public CarStage a_car_exists() {
      wheelStage.$_wheels_exists(4);
      chassisStage.a_chassis_exists();
      return self();
  }

}

//Report
Given a car exists
When the user drives to "London"
Then he gets to "London"

```

If I try this, the inner stages (e.g. `WheelStage`) are "outside" the stage framework and `wheels` would not appear populated anywhere else.

[...]

Issue #203, opened 2016-04-18
Injected stages do not go well together with injected scenario states.


koenusz:
My Situation:
Some steps in your testsuite are usable by many tests. For instance logging into your application. Ideally you would like to program this functionality once and reuse it throughout your testsuite. I found that injecting scenario stages is quite convenient for this purpose. However I found that scenario State is not properly transferred to the parent stage.

An example of the situation:

```
public class MyFeatureSteps extends Stage<MyAppSteps>{	
	
	@ScenarioStage
	LoginSteps loginSteps;
	
	@ExpectedScenarioState
	int newUserId; //The newUserId is not injected here!!
	
	
	public GivenLessomplanSteps a_new_user() {
		loginSteps.a_new_user();
		return self();
	}
	
	public GivenLessomplanSteps the_user_sets_up_my_feature(){
			FeatureFactory.setUp(newUserId);
	}
}	
public Class LoginSteps extends Stage<LoginSteps>{

	@ProvidedScenarioState
	int newUserId;

	public LoginSteps a_new_user() {
		
		newUserId = LoginService.myLoginMethod();	
		return self();
	}		
}
```
The reason for this behavior lies in the way the ValueInjector, which is the class that manages the injected states, gets called. The method readValues is called from a method with the BeforeScenario Annotation. This method is not called when execution is handed over from the child stage to the parent stage thus failing to update the scenario state.

A workaround for this problem is to provide accessors to the state variable and call the get method to update the parent's state.

albertofaci:
I just reported the same thing #202 :)
I'm debugging at the moment. The framework instantiates `loginSteps` for you, but the StepMethodInterceptor would not kick in in when you update `newUserId` and update it in `MyFeatureSteps`

janschaefer:
Yes, this problem is known to me. State transfer currently doesn't work for stages injected into other stages, because the precise semantics is somehow hard to define. This is thus also no documented feature. The point here would be that on calling the injected state the state of the parent stage must be read and the child stage updated. When the call comes back the state of the parent stage must be updated again. For this I somehow have to track the stack of stages. This might be technically possible, however, I wonder whether this is easy to understand.

janschaefer:
As two people had the same intuition, I think this might actually be intuitive :-) I have to think about that a bit, maybe I will create some experimental annotation/feature.

koenusz:
[...]

PR #205, opened 2016-04-19
#203 ( #202 ) implemented ComposedScenarioStage to inject scenario st…

albertofaci:
This implements the `ComposedScenarioStage` annotation, an attempt to be able to inject stages within stages. Tickets #202 and #203.  Just a couple of tweaks in the `ValueInjector` class and job done.

- The composed stages are resolved by type (effectively singletons within the JGiven dependency injection framework). Otherwise we could have variables defined with `@ScenarioState` and bound by name that wouldn't be bound because the containing composed scenarios state are different instances of the same type (it's a bit confusing)

- Nested composed stages are supported and this updates the `stackDepth`.

- The report only displays the top-level stage method call (Although it plays well with `@NestedSteps`)

- `@ScenarioState` fields are updated and can be access from any stage (composed or not)

To solve the example in the ticket

```
public class MyFeatureSteps extends Stage<MyFeatureSteps>{  

    @ComposedScenarioStage
    LoginSteps loginSteps;

    @ExpectedScenarioState
    int newUserId; //The newUserId is now injected here :)


    public GivenLessomplanSteps a_new_user() {
        loginSteps.a_new_user();
        return self();
    }

    public GivenLessomplanSteps the_user_sets_up_my_feature(){
            FeatureFactory.setUp(newUserId);
    }
}   
public Class LoginSteps extends Stage<LoginSteps>{

    @ProvidedScenarioState
    int newUserId;

    public LoginSteps a_new_user() {

        newUserId = LoginService.myLoginMethod();   
        return self();
    }       
}
```

Please let me know if the approach is alright or any comments

### DIFF
--- a/jgiven-core/src/main/java/com/tngtech/jgiven/impl/StandaloneScenarioExecutor.java
+++ b/jgiven-core/src/main/java/com/tngtech/jgiven/impl/StandaloneScenarioExecutor.java
@@ -391,7 +391,7 @@ public class StandaloneScenarioExecutor implements ScenarioExecutor {
     @Override
     @SuppressWarnings( "unchecked" )
     public void injectSteps( Object stage ) {
-        for( Field field : FieldCache.get( stage.getClass() ).getFieldsWithAnnotation( ScenarioStage.class ) ) {
+        for( Field field : FieldCache.get( stage.getClass() ).getFieldsWithAnnotation( ScenarioStage.class, ComposedScenarioStage.class ) ) {
             Object steps = addStage( field.getType() );
             ReflectionUtil.setField( field, stage, steps, ", annotated with @ScenarioStage" );
         }
--- a/jgiven-core/src/main/java/com/tngtech/jgiven/impl/StandaloneScenarioExecutor.java
+++ b/jgiven-core/src/main/java/com/tngtech/jgiven/impl/StandaloneScenarioExecutor.java
@@ -11,7 +11,7 @@ import java.lang.reflect.Method;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
-import java.util.concurrent.atomic.AtomicInteger;
+import java.util.Stack;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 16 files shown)


albertofaci:
Apologies, #202 and #203 are not exactly the same. This only fixes #202 because in #203 the reference happens in the same stage where the composed stage is. I need to investigate a bit more to see if I can solve both in one go. @janschaefer

albertofaci:
I had to remove this bit in `StandaloneScenarioExecutor.update( T t )`. We want to refresh on every call.

```
   if( currentStage == t ) { // NOSONAR: reference comparison OK here
            return t;
        }
```

and then #203 is fixed too

jgiven-core/src/main/java/com/tngtech/jgiven/impl/StandaloneScenarioExecutor.java
```diff
@@ -190,6 +190,8 @@ private void gatherRules( Object stage ) {
 
     private <T> T update( T t ) throws Throwable {
         if( currentStage == t ) { // NOSONAR: reference comparison OK here
+            readScenarioState( currentStage );
```
janschaefer: This will have a big performance overhead. This actually what the if statement was good for: to avoid the overhead of reading and updating the state.

jgiven-core/src/main/java/com/tngtech/jgiven/impl/inject/ValueInjector.java
```diff
@@ -79,6 +84,15 @@ public void readValues( Object object ) {
                 throw new RuntimeException( "Error while reading field " + field, e );
             }
         }
+        for(Field field : getComposedStageFields(object ) ) {
+            try {
+                Object substage = field.get(object);
+                readValues( substage );
```
janschaefer: Ok, I now think I understood your approach. It is a bit different to how I would have implemented it. You essentially treat the state of the substages as part of the parent stage.

janschaefer:
Thank you for the pull request! We have to adress the additional performance overhead somehow.

albertofaci:
No worries. I mean, this approach looked like a quick win, but I don't really know the internals that well. So it's completely up to you

albertofaci:
Hi @janschaefer, have there been any further thoughts about how to integrate this?
Cheers, Alberto

janschaefer:
Hi Alberto,

no, sorry. I didn't have much time lately to look into this.

albertofaci:
Hi, @janschaefer could you please give me a hint on where the performance issue is? I might have a look to see if I can take this further.
Thanks

janschaefer:
The problem is that currently the fields are only read and updated if the current stage actually changes. So all method calls within a stage will not trigger field updates of the stage. Only when the next stage is entered, the fields are read from the current stage and updated in the next stage. As the field read and update is a potentially expensive operation (we go through all fields via reflection) I try to minimize this as far as possible. 
An alternative solution to your solution would be to track when a substage is entered and left. As far as I see it, you have to track the call stack of stages so that you know the calling stage when the substage call has been left.

albertofaci:
@janschaefer I've implemented the stack of calls as you mentioned. I had to add a couple of tweaks to support `@NestedStages`. The limitation is that you can only have a `@ComposedScenarioStage` of each type in every `Stage`, but limitation was already there in the original implementation.

janschaefer:
Is superseeded by #206

PR #206, opened 2016-05-16
Implemented child stages

### LINKED ISSUE #202: Stage composition (stages within stages)

janschaefer:
This is an updated version of PR #205 to fix #202


prathyk:
Hi, @janschaefer ,
when I use stages within stages, the nested stages' steps do not come in the report. Do we have an option to enable that?

janschaefer:
You have to add the `@NestedSteps` annotation to a step method if nested steps should be shown. See http://static.javadoc.io/com.tngtech.jgiven/jgiven-core/0.15.3/com/tngtech/jgiven/annotation/NestedSteps.html for details

PR #207, opened 2016-05-30
don't show multiple indexed attachments when cases are not shown as table

ahus1:
I have a parameterized test case that is not shown as a table (due to different test steps depending on the parameters). 

What I see in HTML5 report of the current version (0.11.3): For every case all attachments of all cases are shown (= multiple paperclips on a single step).

What I want: Only the attachments of the current case should to be shown (without the index). 

This change fixes it for me. I'm not sure if this is the optimal code change as I'm not familiar with AngularJS - but it works for me. Please feel free to refactor.

I'm looking forward to a patch release if possible.


janschaefer:
Fine with me, thanks for the fix!

janschaefer:
I can make a patch release for that.

PR #208, opened 2016-06-18
Allow specifying multiple format annotations on arguments.

### LINKED ISSUE #204: Allow multiple formatters for one property
Hi Jan,

```
public T superbox_query_type_$_with_query_$_is_prepared(@Hidden Class<? extends SuperBoxController> superboxClass,
			JAFQueryType queryType, @Quoted @NotSetIfNull String query) {
```
throws a

com.tngtech.jgiven.exception.AmbiguousFormattingException.

Is there any reason why multiple formatters can't be used in the order they are declared? To achieve that behaviour, I need to provide another formatter annotation that combines both formatters which results in increased code verbosity.

Best regards,
Niko

Airblader:
fixes #204


Airblader:
(This is not meant to be merged yet, it's just easier to talk about these changes this way)

jgiven-core/src/main/java/com/tngtech/jgiven/exception/AmbiguousFormattingException.java
```diff
@@ -1,12 +0,0 @@
-package com.tngtech.jgiven.exception;
```
Airblader: Depending on how strongly you feel about backwards compatibility, we would have to keep this class even if now unused.

janschaefer: It is fine to remove the class. No existing code should rely on its existence

janschaefer:
Looks good! As discussed in ticket #204 it would be great if you could catch the ClassCastException, and turn it into a JGivenWrongUsageException (or subclass of that) and add a little test for that. You could also already add an entry to the CHANGELOG.md

Airblader:
@janschaefer Done!

janschaefer:
Great!

Issue #50, opened 2014-12-20
Allow for commenting of steps


janschaefer:
It is already possible to provide an extended description to steps by using the `@ExtendedDescription` annotation. However, sometimes you want to comment a certain step-invocation and not all invocations of that step. This comment should also appear in the report.

Example:
```
given().something();
when().some_step()
  .and().some_surprising_step()
   .comment("a comment is especially important here because the step surprises the reader")
  then().something();
```

Airblader:
Just to clarify, this should be really a new piece of information on a step rather than just setting the extendedDescription of a particular step instance?

Airblader:
Also, from my understanding of the codebase, this would roughly involve the following:

* Add a new `comment` property to `Word.java` and add it to the JSON protocol.
* Introduce a new `comment` method in `StageBase.java` (or `Stage.java` to make it language-aware).
 * Introduce a new annotation for such a method, similar to `@IntroWord`.
 * Handle this annotation in `stepMethodInvoked#stepMethodInvoked` by setting the `comment` property of `currentStep` to the argument of the called method (TODO: can we somehow enforce that such a comment method must have exactly one argument of type String?)
* Handle the `comment` property in the HTML reporter (and possibly others?)
 * TODO: how should it be displayed?

I hope I got the basic design of this feature right.

PR #210, opened 2016-06-25
Allow providing an additional comment on step method invocations

Airblader:
This PR is not yet complete. As of now, it implements the basic mechanism and printing a comment in the PlainTextReporter. Missing:

* Displaying it in the HTML report.
* Tests, obviously. :-)

I just want to run it by you first to get feedback on the overall solution and also to discuss how the HTML report should display the comment.

### DIFF
--- a/jgiven-core/src/main/java/com/tngtech/jgiven/impl/ScenarioModelBuilder.java
+++ b/jgiven-core/src/main/java/com/tngtech/jgiven/impl/ScenarioModelBuilder.java
@@ -141,6 +141,24 @@ public class ScenarioModelBuilder implements ScenarioListener {
         introWord.setValue( value );
     }
 
+    @Override
+    public void stepCommentAdded( List<NamedArgument> arguments ) {
+        if( arguments == null || arguments.size() != 1 ) {
+            throw new JGivenWrongUsageException( "A step comment method must have exactly one parameter." );
+        }
+
+        if( !( arguments.get( 0 ).getValue() instanceof String ) ) {
+            throw new JGivenWrongUsageException( "The step comment method parameter must be a string." );
+        }
+
+        if( currentStep == null ) {
+            throw new JGivenWrongUsageException( "A step comment must be added after the corresponding step, "
+                    + "but no step has been executed yet." );
+        }
+
+        currentStep.setComment( (String) arguments.get( 0 ).getValue() );
+    }
+
     private ScenarioCaseModel getCurrentScenarioCase() {
         if( scenarioCaseModel == null ) {
             scenarioStarted( "A Scenario" );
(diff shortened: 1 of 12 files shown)


jgiven-core/src/main/java/com/tngtech/jgiven/impl/ScenarioModelBuilder.java
```diff
@@ -141,6 +141,24 @@ public void introWordAdded( String value ) {
...
     }
 
+    @Override
+    public void stepCommentAdded( List<NamedArgument> arguments ) {
+        if( arguments == null || arguments.size() != 1 ) {
+            throw new JGivenWrongUsageException( "A step comment method must have exactly one parameter." );
+        }
+
+        if( !( arguments.get( 0 ).getValue() instanceof String ) ) {
+            throw new JGivenWrongUsageException( "The step comment method parameter must be a string." );
```
janschaefer: One could argue here that we could use any object and just use the toString method

janschaefer:
Looks really good so far. The big question is now how to visualize this in the HTML report. One possibility is to show it like the ExtendedDescription with a tooltip. If there is also an extended description that might be combined in a single tooltip, maybe.

janschaefer:
this PR implements #50

Airblader:
@janschaefer I was just wondering how this plays together with `CurrentStep`. Why is `addAttachment` exposed via `CurrentStep`, but `comment` is a direct stage method? Should we reconsider this design?

Also: is there an easy(!) way to generate the HTML report for the tests? :-) I can't seem to get the classpath right when calling it from command-line; when simply calling the generator through a test, I get a NPE in `unzipApp`. :-(

jgiven-core/src/main/java/com/tngtech/jgiven/impl/ScenarioModelBuilder.java
Airblader: One could indeed. But is there a usecase for it? :-) Seems like a non-string comment is actually more of an attachment.

janschaefer:
As the comment is something which is a feature for ad-hoc commenting steps I think it makes more sense to have that as a Stage method. 

Regarding the test problem. It should work when you add the folder `jgiven-html5-report/build/resources/main` to the test classpath

Airblader:
> As the comment is something which is a feature for ad-hoc commenting steps I think it makes more sense to have that as a Stage method.

Alright, let's proceed then. I managed to hack together a test that creates the HTML output now, so I'll be working on the tooltip and tests.

Airblader:
Here's what it'll look like:

screenshot-2016-06-26_23-57-40
screenshot-2016-06-26_23-57-37
screenshot-2016-06-26_23-57-34

janschaefer:
Looks good :-)

janschaefer:
If you could add a test (does not have to be on the HTML level) we can merge it

Airblader:
@janschaefer And done!

janschaefer:
Thanks!

PR #213, opened 2016-07-01
Allow adding tags dynamically.

### LINKED ISSUE #172: Adding tags dynamically
It might sometimes be useful to add tags dynamically to scenarios

Airblader:
This commit allows adding tags dynamically to a scenario at runtime.
Furthermore, it implicitly fixes an issue where nested tags, description
and href would not be evaluated if ignoreValue was set to true.

fixes #172

### DIFF
--- a/jgiven-core/src/main/java/com/tngtech/jgiven/impl/ScenarioModelBuilder.java
+++ b/jgiven-core/src/main/java/com/tngtech/jgiven/impl/ScenarioModelBuilder.java
@@ -1,8 +1,10 @@
 package com.tngtech.jgiven.impl;
 
 import com.google.common.base.CaseFormat;
+import com.google.common.base.Optional;
 import com.google.common.base.Strings;
 import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Iterables;
 import com.google.common.collect.Lists;
 import com.tngtech.jgiven.annotation.*;
 import com.tngtech.jgiven.attachment.Attachment;
@@ -381,26 +383,26 @@ public class ScenarioModelBuilder implements ScenarioListener {
 
     public void addTags( Annotation... annotations ) {
         for( Annotation annotation : annotations ) {
-            List<Tag> tags = toTags( annotation );
-            this.reportModel.addTags( tags );
-            this.scenarioModel.addTags( tags );
+            addTags( toTags( annotation ) );
         }
+    }
+
+    private void addTags( List<Tag> tags ) {
+        this.reportModel.addTags( tags );
+        this.scenarioModel.addTags( tags );
     }
 
     public List<Tag> toTags( Annotation annotation ) {
         Class<? extends Annotation> annotationType = annotation.annotationType();
-        IsTag isTag = annotationType.getAnnotation( IsTag.class );
-        TagConfiguration tagConfig;
-        if( isTag != null ) {
-            tagConfig = fromIsTag( isTag, annotation );
-        } else {
-            tagConfig = configuration.getTagConfiguration( annotationType );
-        }
-
+        TagConfiguration tagConfig = toTagConfiguration( annotationType );
         if( tagConfig == null ) {
             return Collections.emptyList();
         }
 
+        return toTags( tagConfig, Optional.of( annotation ) );
+    }
+
+    private List<Tag> toTags( TagConfiguration tagConfig, Optional<Annotation> annotation ) {
         Tag tag = new Tag( tagConfig.getAnnotationType() );
 
         if( !Strings.isNullOrEmpty( tagConfig.getName() ) ) {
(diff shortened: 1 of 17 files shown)


jgiven-core/src/main/java/com/tngtech/jgiven/impl/ScenarioModelBuilder.java
```diff
@@ -382,26 +384,26 @@ private void readAnnotations( Class<?> testClass, Method method ) {
...
-            List<Tag> tags = toTags( annotation );
-            this.reportModel.addTags( tags );
-            this.scenarioModel.addTags( tags );
+            addTags( toTags( annotation ) );
         }
     }
 
+    private void addTags( List<Tag> tags ) {
+        this.reportModel.addTags( tags );
+        this.scenarioModel.addTags( tags );
```
janschaefer: why do you add the tags to the reportModel and the scenarioModel?

Airblader: Because the normal path of adding tags via annotation does the same. Which one shouldn't it be added to?

Airblader:
Rebased to the current master.

jgiven-core/src/main/java/com/tngtech/jgiven/impl/ScenarioModelBuilder.java
janschaefer: Then it should be correct :-D

Airblader: OK. :-) Then everything is ready from my side.

janschaefer:
Cool! I still wonder, however, whether it wouldn't be useful to have a method in the CurrentStep class to add tags to the scenario. What do you think?

Airblader:
Hm, I'm not sure. `CurrentStep` allows access to the current step method, but tags are related to entire scenarios, not individual steps. So semantically it doesn't seem to belong into `CurrentStep` to me. Is there any situation where `CurrentStep` can be used, but `getScenario()` can't?

An alternative would be to expose `CurrentStep#getScenario()` which would then also allow adding tags from there. At the same time I feel that it proves my point about getting to scenario-wide information from step-specific mechanisms… :-)

janschaefer:
Yes tags belong to the Scenario and not the step. Maybe one could introduce a new interface CurrentScenario that is returned by CurrentStep#getScenario()? CurrentScenario could have a method addTag then.

Airblader:
@janschaefer OK, sure. Should we remove `ScenarioBase#addTag` again, then?

Airblader:
Small addendum:

> that is returned by CurrentStep#getScenario()?

Is that necessary? I still think getting to the scenario from a step sounds a little off. If we have a `CurrentScenario` interface, the user can just inject this directly anyway. No need to go through `CurrentStep`, right?

janschaefer:
Also true and makes sense

janschaefer:
I am still not quite sure about the use cases of this feature. :-)

Airblader:
> Also true and makes sense

Is this also a "yes" to removing it from ScenarioBase again? :-) The reason I'm asking is because for consistency we should probably do the same for the `section` method in ScenarioBase.

Airblader:
> I am still not quite sure about the use cases of this feature. :-)

Since you opened the ticket I assumed it had a valid usecase. But of course let's not merge anything if we can't verify that it's useful. Nobody wants bloat. :-)

Do tags work if stage methods are annotated with them rather than the scenario? I feel like the "major" usecase here is that the user doesn't need to think about tags for every new test they write, but instead one could annotate the test method to include a certain tag. That doesn't require programmatic tags, however; simply annotating the stage method would suffice for that.

As for a need for actual programmatic tags, I *could* think of adding a tag based on an argument provided to a stage method. Although one could probably argue that it'd be worth having separate stage methods for such cases.

Airblader:
Programmatic tags could also be useful if the behavior of a stage method depends on the scenario state. Consider a scenario

Airblader:
Programmatic tags could also be useful if the behavior of a stage method depends on the scenario state. Consider a `When` method like the following

````
public SELF the_order_is_submitted() {
    currentScenario.addTag( Order.class, paymentMethod );
    // …
}
```

Or a `Given` method like this:

````
public SELF the_backend_responds_with( @FormatErrorCode BackendErrorCode error ) {
    currentScenario.addTag( BackendFailure.class, error );
    // …
}
```

Airblader:
I actually really like the idea of having stage methods provide tags rather than scenarios, at least for some cases. I think independent of the discussion about programmatic access, it would be useful to be able to define tags on stage method.

Airblader:
So just to summarize on what we discussed today:

* We want to allow tag annotations on stage methods and classes.
* We want to use a `CurrentScenario` interface similar to `CurrentStep`.
* We want to move `section()` and `addTag()` to said interface.
* We want to allow dynamic tags through `CurrentScenario#addTag`.

Airblader:
I'm actually not sure that we should move `addSection`. We'd also have to remove `section()` from ScenarioTestBase, but I think that `section()` should be in there just like `given()` / `when()` / `then()` because all of those methods are used in a way where their order is relevant to both the output and the way the scenario reads. So I'll skip this for now.

janschaefer:
I agree :-)

jgiven-core/src/main/java/com/tngtech/jgiven/impl/StandaloneScenarioExecutor.java
```diff
@@ -106,6 +108,15 @@ public void setExtendedDescription( String extendedDescription ) {
         }
     }
 
+    class ScenarioAccessImpl implements CurrentScenario {
```
janschaefer: should be static, right?

jgiven-core/src/main/java/com/tngtech/jgiven/impl/ScenarioModelBuilder.java
```diff
@@ -172,6 +174,9 @@ public void stepMethodInvoked( Method method, List<NamedArgument> arguments, Inv
         } else if( method.isAnnotationPresent( StepComment.class ) ) {
             stepCommentAdded( arguments );
         } else {
+            addTags( method.getAnnotations() );
+            addTags( method.getDeclaringClass().getAnnotations() );
```
Airblader: I tried passing the receiver from cglib instead of using `getDeclaringClass`, but cglib subclasses the class to create the proxy which means any annotation not tagged with `@Inherited` gets lost. I'm not really sure if either way would be better than the other, so maybe this is completely fine…?

Airblader:
@janschaefer OK, good to go! See my one comment above.

janschaefer:
Cool, thanks!

PR #215, opened 2016-07-11
Extend @As with a provider mechanism.

### LINKED ISSUE #189: Format step method dynamically without parameters
Hi Jan,

we've implemented some steps that are within some date ranges like this:

```
Given logged in user is system
     And  today is 2016-02-18
     And  response for rfqquestionnaire that is submittable by suma.contact@sobis.com
     And corresponding RfQ
    When start date is set to 2015-08-18 and due date is set to 2016-05-18
     And question changes
    Then no errors are raised
```

To be able to render the second (today) line, I need to supply the current date as a parameter, but I prefer to have a helper method like today() _without having to supply a parameter_. With @NestedSteps, this could be wrapped a little bit, but it is only a workaround. So basically I have to be able to have something similar to the CaseDescriptionProvider for step method that allows me to to some coding.

Is this already possible somehow?

Best regards,
Niko

Airblader:
This change extends the @As annotation for step / stage methods and classes
with an AsProvider provider() method to allow a more flexible way of
customizing the result.

fixes #189

### DIFF
--- a/jgiven-core/src/main/java/com/tngtech/jgiven/impl/ScenarioModelBuilder.java
+++ b/jgiven-core/src/main/java/com/tngtech/jgiven/impl/ScenarioModelBuilder.java
@@ -1,5 +1,17 @@
 package com.tngtech.jgiven.impl;
 
+import java.lang.annotation.Annotation;
+import java.lang.reflect.Method;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import java.util.Set;
+import java.util.Stack;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import com.google.common.base.CaseFormat;
 import com.google.common.base.Optional;
 import com.google.common.base.Strings;
@@ -21,12 +33,6 @@ import com.tngtech.jgiven.impl.util.AssertionUtil;
 import com.tngtech.jgiven.impl.util.ReflectionUtil;
 import com.tngtech.jgiven.impl.util.WordUtil;
 import com.tngtech.jgiven.report.model.*;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
-import java.lang.annotation.Annotation;
-import java.lang.reflect.Method;
-import java.util.*;
 
 public class ScenarioModelBuilder implements ScenarioListener {
     private static final Logger log = LoggerFactory.getLogger( ScenarioModelBuilder.class );
@@ -212,7 +218,8 @@ public class ScenarioModelBuilder implements ScenarioListener {
         }
         As as = paramMethod.getAnnotation( As.class );
         if( as != null ) {
-            return as.value();
+            AsProvider provider = ReflectionUtil.newInstance( as.provider() );
+            return provider.as( as, paramMethod );
         }
 
         return nameWithSpaces( paramMethod );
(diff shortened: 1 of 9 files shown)


Airblader:
Obviously missing things:

- [ ] Documentation
- [ ] Tests

@janschaefer I liked having two separate `as()` methods for the method versus class cases better. In fact, we could we even differentiate between calling this for a step method and for a test method, since the annotation works on both. Should we?

Airblader:
@janschaefer OK, you can take a look at it now.

jgiven-core/src/main/java/com/tngtech/jgiven/impl/params/DefaultAsProvider.java
```diff
@@ -0,0 +1,56 @@
...
+     * Returns the value of the {@link As} annotation if a value was specified.
+     * Otherwise, this transforms the method name into a readable sentence and returns it.
+     */
+    @Override
+    public String as( As annotation, Method method ) {
+        if( annotation != null && annotationHasExplicitValue( annotation ) ) {
+            return annotation.value();
+        }
+
+        String methodName = method.getName();
```
Airblader: So I think one can make either argument for or against having this in the DefaultAsProvider. My rationale is that this way, a user can extend from this provider and utilize this, e.g., in the spirit of the original motivation for #189:

````
@As( provider = ReplaceTodayProvider.class )
public SELF a_request_with_request_date_TODAY() { … }

public static class ReplaceTodayProvider extends DefaultAsProvider {
    @Override
    public String as( As annotation, Method method ) {
        String words = super.as( annotation, method );
        return String.replace( "TODAY", formatDate( new Date() ) );
    }
}
```

jgiven-core/src/main/java/com/tngtech/jgiven/impl/ScenarioModelBuilder.java
```diff
@@ -210,12 +208,11 @@ private String getDescription( Method paramMethod ) {
...
         }
-        As as = paramMethod.getAnnotation( As.class );
-        if( as != null ) {
-            return as.value();
-        }
 
-        return nameWithSpaces( paramMethod );
+        As as = paramMethod.getAnnotation( As.class );
+        return ReflectionUtil
+            .newInstance( as == null ? DefaultAsProvider.class : as.provider() )
```
janschaefer: I would prefer to create the DefaultAsProvider without using reflection

jgiven-core/src/main/java/com/tngtech/jgiven/impl/params/DefaultAsProvider.java
```diff
@@ -0,0 +1,56 @@
...
+
+import com.tngtech.jgiven.annotation.As;
+import com.tngtech.jgiven.annotation.AsProvider;
+import com.tngtech.jgiven.config.AbstractJGivenConfiguration;
+import com.tngtech.jgiven.config.ConfigurationUtil;
+import com.tngtech.jgiven.impl.util.WordUtil;
+
+/**
+ * The default provider for a stage method, scenario or scenario class.
+ *
```
janschaefer: @since 0.12.0

jgiven-core/src/main/java/com/tngtech/jgiven/impl/ScenarioModelBuilder.java
Airblader: Yeah, that makes sense, given that it will happen for pretty much all methods.

Airblader:
@janschaefer Updated.

PR #230, opened 2016-08-25
upload of new asciidocs

CPTblackadder:
markdown documentation ported over to asciidocs


janschaefer:
Cool, thanks. Can you check the build? The examples seem not to compile anymore. You can also check the Travis output.

jgiven-examples/src/test/java/com/tngtech/jgiven/examples/webdriver/WebDriver.java
```diff
@@ -0,0 +1,10 @@
+package com.tngtech.jgiven.examples.webdriver;
+
+public class WebDriver {
```
janschaefer: maybe you could add the missing dependency to the build.gradle instead? Would also result in the correct package then:

org.seleniumhq.selenium:htmlunit-driver:2.21

janschaefer:
Documentation is published :-)

http://jgiven.org/userguide/

PR #231, opened 2016-08-30
Upgrade to Gradle 3.0

janschaefer:


Airblader:
So JGiven is now dropping support for Java 6?

janschaefer:
Hehe, no. It will still be built for Java 6, but not using Java 6 for building. This introduces a small risk of accidentally using a non-JDK6 API, but as long as I configure IntelliJ correctly, this shouldn't happen either.

Airblader:
OK. :-)

ahus1:
So be extra sure that you use only Java 6 APIs (not only Java 6 Syntax) have a look at http://www.mojohaus.org/animal-sniffer/.

janschaefer:
Thanks @ahus1 for the hint. I will have a look at that!

janschaefer:
Thanks again @ahus1, works like a charm! :-)

PR #240, opened 2016-10-09
create files for all attachments

### LINKED ISSUE #228: Attachment in JSON format in not written to a file
If you define an attachment with MediaType JSON_UTF_8, no file is created by the HTML5 report. 

Html5AttachmentGenerator will create only files for "text/\*" and "image/\*" types.

I would expect to have a file regardless media type (maybe with a default extension of "bin" like binary). For "application/json" the extension should be ".json". 

Please comment if this is what other would expect as well. Thanks!

ahus1:
Instead of creating files for text and image media types, this now creates attachments for all media types.

The the binary vs. text content will be determined by the _binary_ flag of the attachment. The file extension will be determined by the media type.

A new test in Html5AttachmentGeneratorTest ensure that both text and binary files are created.

Closes #228

### DIFF
--- a/jgiven-html5-report/src/test/java/com/tngtech/jgiven/report/html5/Html5AttachmentGeneratorTest.java
+++ b/jgiven-html5-report/src/test/java/com/tngtech/jgiven/report/html5/Html5AttachmentGeneratorTest.java
@@ -2,9 +2,24 @@ package com.tngtech.jgiven.report.html5;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
+import com.tngtech.jgiven.attachment.Attachment;
+import com.tngtech.jgiven.attachment.MediaType;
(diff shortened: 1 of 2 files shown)


jgiven-html5-report/src/test/java/com/tngtech/jgiven/report/html5/Html5AttachmentGeneratorTest.java
```diff
@@ -17,4 +32,37 @@ public void testFileNameGeneration() {
         assertThat( generator.getTargetFile( "foo", "png" ).getName() ).isEqualTo( "foo5.png" );
 
     }
+
+    @Test
+    public void testFileCreation() throws IOException {
```
janschaefer: I would have preferred a JGiven test :-)

PR #241, opened 2016-10-09
allow `gradlew publishToMavenLocal`

ahus1:
As a contributor I want to use `gradlew publishToMavenLocal` so I can test my contributions easily together with my local projects.

Therefore I added _maven-publish_. As I'm unexperienced with gradle, I might have missed something. But it now works for me.

Please advise if changes are necessary.


janschaefer:
I think that is fine and might be useful for me as well :-)

PR #242, opened 2016-10-09
use full class name as ID of a tag

ahus1:
In our project we use very many tags, and we order them in hierarchies and with inner classes. Sometimes inner classes have the same names (as they are the leaves of our tree).  This is currently not supported by JGiven, as it uses the class' simpleName to index the tags. I know that this is to some extent documented in @IsTag#name - but it seems to be a bit outdated as it talks about "type name".

This tries to introduce the full class name to derive the IDs of the tags, while preserving the existing functionality to use the simple name as the default name of the tag. The ID will only be used for the internal indexing, and will not be shown to the user.

I've tested this with the HTML5 report and it works for me AFAICS. Let me know what you think - I wonder if it breaks functionality or (too much) backwards compatibility.

In addition to this the _...#/tag/name_ navigation used in the HTML5 report should use something like _...#/tagid/id_ to uniquely identify tags not by their name, but by their ID. I'll look into this once we agree on the Java code.

### DIFF
--- a/jgiven-core/src/main/java/com/tngtech/jgiven/config/TagConfiguration.java
+++ b/jgiven-core/src/main/java/com/tngtech/jgiven/config/TagConfiguration.java
@@ -16,6 +16,7 @@ import java.util.List;
  */
 public class TagConfiguration {
     private final String annotationType;
+    private final String annotationFullType;
     private boolean ignoreValue;
     private boolean explodeArray = true;
     private boolean prependType;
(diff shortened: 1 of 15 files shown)


janschaefer:
I haven't tested yet, but I wonder how this works without actually adapting the HTML report :-)

janschaefer:
I think the change is good and I also agree with your suggestion of introducing a `tagid/id` navigation

ahus1:
The javascript changes have finally arrived for this PR. the `/tagid` URL fragment is now used when selecting a tag from the navigation on the left (as long as it is not a name based tag), or when clicking on a tag next to a test case.

janschaefer:
Looks good. Thank you!

PR #245, opened 2016-10-17
Add initial version of JGiven Gradle plugin

### LINKED ISSUE #98: Gradle report plugin
Like there is already a maven plugin a corresponding gradle plugin would be nice.

wolfs:
This would fix #98.

The plugin currently adds a `jgiven<Test Task name>Report` task
 for each test task and generates the report (if JGiven JSON files have been generated)
 to `build/reports/jgiven/<Test Task Name>`.

Things which still need some work:

 - Add the possibility to generate other reports apart from the html
 report
 - Add some more tests regarding up to date checking
 - Should it be possible to generate one report for all tests executed
   in
   - a multimodule build
   - different test tasks
 - Use the built version of JGiven in the tests instead of the last
   released version
 - Polish the code

### DIFF
--- a/jgiven-core/src/main/java/com/tngtech/jgiven/impl/Config.java
+++ b/jgiven-core/src/main/java/com/tngtech/jgiven/impl/Config.java
@@ -1,12 +1,11 @@
 package com.tngtech.jgiven.impl;
 
-import java.io.File;
-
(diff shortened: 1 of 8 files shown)


janschaefer:
Awesome! Thanks a lot!

PR #246, opened 2016-10-26
Add configurable reports for the Gradle plugin

wolfs:
The html and text reports can now be configured.


janschaefer:
One of your tests fails :-)

wolfs:
Fixed.

Issue #229, opened 2016-08-23
JGiven does not work with Spring JUnit Rules


janschaefer:
Since v4.2, Spring provides a SpringClassRule and a SpringMethodRule. When using these rules instead of the SpringRunner, JGiven throws a `JGivenInternalDefectException`

PR #250, opened 2016-11-11
Use Spring's @Rule support instead of @RunWith(SpringJunit4ClassRunner)

catchin:
Use Spring's @Rule support added in v4.2 for the SpringScenarioTest instead of using SpringJUnit4ClassRunner.

Enables the use of custom JUnit runners like DataProviderRunner (see example test).

### DIFF
--- a/jgiven-examples/src/test/java/com/tngtech/jgiven/examples/pancakes/test/SpringPanCakeScenarioTest.java
+++ b/jgiven-examples/src/test/java/com/tngtech/jgiven/examples/pancakes/test/SpringPanCakeScenarioTest.java
@@ -1,30 +1,18 @@
 package com.tngtech.jgiven.examples.pancakes.test;
 
-import org.junit.Before;
 import org.junit.Test;
 import org.junit.runner.RunWith;
-import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.beans.factory.config.AutowireCapableBeanFactory;
 import org.springframework.test.context.ContextConfiguration;
 import org.springframework.test.context.junit4.SpringJUnit4ClassRunner;
 
-import com.tngtech.jgiven.examples.pancakes.app.SpringConfig;
 import com.tngtech.jgiven.examples.pancakes.test.steps.GivenIngredients;
 import com.tngtech.jgiven.examples.pancakes.test.steps.ThenMeal;
 import com.tngtech.jgiven.examples.pancakes.test.steps.WhenCook;
-import com.tngtech.jgiven.integration.spring.SpringCanWire;
-import com.tngtech.jgiven.junit.ScenarioTest;
+import com.tngtech.jgiven.integration.spring.SpringScenarioTest;
 
 @RunWith( SpringJUnit4ClassRunner.class )
-@ContextConfiguration( classes = SpringConfig.class )
-public class SpringPanCakeScenarioTest extends ScenarioTest<GivenIngredients, WhenCook, ThenMeal> {
-    @Autowired
-    private AutowireCapableBeanFactory beanFactory;
-
-    @Before
-    public void setupSpring() {
-        wireSteps( new SpringCanWire( beanFactory ) );
-    }
+@ContextConfiguration( classes = TestSpringConfig.class )
+public class SpringPanCakeScenarioTest extends SpringScenarioTest<GivenIngredients, WhenCook, ThenMeal> {
 
     @Test
     public void a_pancake_can_be_fried_out_of_an_egg_milk_and_flour() {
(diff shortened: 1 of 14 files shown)


jgiven-spring/src/main/java/com/tngtech/jgiven/integration/spring/SpringScenarioTest.java
```diff
@@ -19,10 +16,16 @@
  *
  * @since 0.8.0
  */
-public class SpringScenarioTest<GIVEN, WHEN, THEN> extends
-        ScenarioTest<GIVEN, WHEN, THEN> implements BeanFactoryAware {
+public class SpringScenarioTest<GIVEN, WHEN, THEN> extends InternalSpringScenarioTest<GIVEN, WHEN, THEN> {
```
janschaefer: I think we need to introduce a new class here, something like `RuleBasedSpringScenarioTest`. Otherwise users of old Spring versions will not be able to use that class anymore.

jgiven-spring/src/test/java/com/tngtech/jgiven/integration/spring/test/XmlConfiguredSpringScenarioTestTest.java
```diff
@@ -1,13 +1,10 @@
 package com.tngtech.jgiven.integration.spring.test;
 
 import org.junit.Test;
-import org.junit.runner.RunWith;
 import org.springframework.test.context.ContextConfiguration;
-import org.springframework.test.context.junit4.SpringJUnit4ClassRunner;
 
 import com.tngtech.jgiven.integration.spring.SpringScenarioTest;
 
-@RunWith( SpringJUnit4ClassRunner.class )
```
janschaefer: Same as above: we still need to support the runner-based execution

jgiven-spring/src/main/java/com/tngtech/jgiven/integration/spring/InternalSpringScenarioTest.java
```diff
@@ -0,0 +1,38 @@
...
+ * {@link SpringMethodRule}s validations, but should not be used directly.
+ * Instead, use {@link SpringScenarioTest}.
+ *
+ * @param <GIVEN>
+ * @param <WHEN>
+ * @param <THEN>
+ *
+ * @since 0.14.0
+ */
+public abstract class InternalSpringScenarioTest<GIVEN, WHEN, THEN> extends ScenarioTestBase<GIVEN, WHEN, THEN> implements BeanFactoryAware {
```
janschaefer: As the constructor is already package-private you could also make the class itself package-private

jgiven-examples/src/test/java/com/tngtech/jgiven/examples/pancakes/test/SpringPanCakeScenarioTest.java
```diff
@@ -1,30 +1,15 @@
...
-@ContextConfiguration( classes = SpringConfig.class )
-public class SpringPanCakeScenarioTest extends ScenarioTest<GivenIngredients, WhenCook, ThenMeal> {
-    @Autowired
-    private AutowireCapableBeanFactory beanFactory;
-
-    @Before
-    public void setupSpring() {
-        wireSteps( new SpringCanWire( beanFactory ) );
-    }
+@ContextConfiguration( classes = TestSpringConfig.class )
```
janschaefer: good catch, thanks for adapting that!

jgiven-spring/src/main/java/com/tngtech/jgiven/integration/spring/InternalSpringScenarioTest.java
catchin: No, this does not work because of Springs ClassRule:
`org.junit.internal.runners.rules.ValidationError: The @ClassRule 'springClassRule' must be declared in a public class.`
Would you think it's more consistent to have the constructor public then?

catchin: And move it to a package named "internal"?

catchin:
I created new classes as suggested.

jgiven-spring/src/main/java/com/tngtech/jgiven/integration/spring/InternalSpringScenarioTest.java
janschaefer: no I think that would be overkill. The class is visible in any case, so just leave it as it is

catchin:
Or SimpleSpringRuleScenarioTest and SpringRuleScenarioTest ?

janschaefer:
That is even better :-)

catchin:
Incorporated the changes.

catchin:
fixes #229

PR #261, opened 2016-12-11
Rewrite of the Spring integration

### LINKED ISSUE #259: Nested steps are not reported with jgiven-spring
Hi

When running a [spring scenario test](https://github.com/TNG/JGiven/tree/master/examples/spring-boot/src/test/java/com/tngtech/jgiven/example/springboot) with  ```@NestedSteps``` , it seems that nested steps are not reported. 

Did I miss something in my config ?

Thanks for your help

janschaefer:
- Instead of using a Spring-based interceptor, the new integration uses
  the standard JGiven interceptor using ByteBuddy.
- The implementation is based on using a BeanFactoryPostProcessor that
  replaces the original Stage bean class with the one created by
  ByteBuddy

Fixes #259


jgiven-spring/src/main/java/com/tngtech/jgiven/integration/spring/JGivenBeanFactoryPostProcessor.java
```diff
@@ -0,0 +1,46 @@
...
+import java.util.Iterator;
+
+
+public class JGivenBeanFactoryPostProcessor implements BeanFactoryPostProcessor {
+
+    private final ByteBuddyStageCreator buddyStageCreator = new ByteBuddyStageCreator();
+
+    @Override
+    public void postProcessBeanFactory(ConfigurableListableBeanFactory beanFactory) throws BeansException {
+        Iterator<String> beanNames = beanFactory.getBeanNamesIterator();
```
janschaefer: Good point! I changed to `getBeanDefinitionNames`

PR #264, opened 2016-12-21
Added required attribute to ScenarioState.

### LINKED ISSUE #255: Introduce mandatory Scenario State
It would be useful to have an attribute `mandatory = true` in the `@ScenarioState` annotation. If this is set, JGiven will fail if no such `@ScenarioState` would have been provided.

Example:
```
class MyStage {
    @ScenarioState(mandatory = true)
    String mandatoryState;
}
```

Airblader:
This commit adds a new boolean attribute "required" to both
ScenarioState and ExpectedScenarioState. If set to true on a
field within a stage, corresponding tests will fail automatically
if the state hasn't been provided.

fixes #255

### DIFF
--- a/jgiven-core/src/main/java/com/tngtech/jgiven/impl/inject/ValueInjector.java
+++ b/jgiven-core/src/main/java/com/tngtech/jgiven/impl/inject/ValueInjector.java
@@ -1,6 +1,5 @@
 package com.tngtech.jgiven.impl.inject;
 
-import java.lang.annotation.Annotation;
 import java.lang.reflect.Field;
 import java.util.List;
 import java.util.Map;
@@ -9,12 +8,14 @@ import java.util.concurrent.ConcurrentHashMap;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
 import com.tngtech.jgiven.annotation.ExpectedScenarioState;
 import com.tngtech.jgiven.annotation.ProvidedScenarioState;
 import com.tngtech.jgiven.annotation.ScenarioState;
 import com.tngtech.jgiven.annotation.ScenarioState.Resolution;
 import com.tngtech.jgiven.exception.AmbiguousResolutionException;
+import com.tngtech.jgiven.exception.JGivenMissingRequiredScenarioStateException;
 import com.tngtech.jgiven.impl.util.FieldCache;
 
 /**
@@ -42,41 +43,45 @@ public class ValueInjector {
 
         Map<Object, Field> resolvedFields = Maps.newHashMap();
 
-        for( Field field : getScenarioFields( object ) ) {
-            field.setAccessible( true );
-            Resolution resolution = getResolution( field );
+        for( ScenarioStateField field : getScenarioFields( object ) ) {
+            field.getField().setAccessible( true );
+            Resolution resolution = field.getResolution();
             Object key = null;
             if( resolution == Resolution.NAME ) {
-                key = field.getName();
+                key = field.getField().getName();
             } else {
-                key = field.getType();
+                key = field.getField().getType();
             }
             if( resolvedFields.containsKey( key ) ) {
                 Field existingField = resolvedFields.get( key );
                 throw new AmbiguousResolutionException( "Ambiguous fields with same " + resolution + " detected. Field 1: " +
-                        existingField + ", field 2: " + field );
+                        existingField + ", field 2: " + field.getField() );
             }
-            resolvedFields.put( key, field );
+            resolvedFields.put( key, field.getField() );
         }
 
         validatedClasses.put( object.getClass(), Boolean.TRUE );
     }
 
-    private List<Field> getScenarioFields( Object object ) {
-        return FieldCache.get( object.getClass() ).getFieldsWithAnnotation( ScenarioState.class, ProvidedScenarioState.class,
-            ExpectedScenarioState.class );
+    private List<ScenarioStateField> getScenarioFields( Object object ) {
+        @SuppressWarnings( "unchecked" )
+        List<Field> scenarioFields = FieldCache
+            .get( object.getClass() )
+            .getFieldsWithAnnotation( ScenarioState.class, ProvidedScenarioState.class, ExpectedScenarioState.class );
+
+        return Lists.transform( scenarioFields, ScenarioStateField.fromField );
     }
 
     @SuppressWarnings( "unchecked" )
     public void readValues( Object object ) {
         validateFields( object );
-        for( Field field : getScenarioFields( object ) ) {
+        for( ScenarioStateField field : getScenarioFields( object ) ) {
             try {
-                Object value = field.get( object );
+                Object value = field.getField().get( object );
                 updateValue( field, value );
-                log.debug( "Reading value {} from field {}", value, field );
+                log.debug( "Reading value {} from field {}", value, field.getField() );
             } catch( IllegalAccessException e ) {
-                throw new RuntimeException( "Error while reading field " + field, e );
+                throw new RuntimeException( "Error while reading field " + field.getField(), e );
             }
         }
     }
@@ -84,15 +89,19 @@ public class ValueInjector {
     @SuppressWarnings( "unchecked" )
     public void updateValues( Object object ) {
         validateFields( object );
-        for( Field field : getScenarioFields( object ) ) {
-            try {
-                Object value = getValue( field );
-                if( value != null ) {
-                    field.set( object, value );
-                    log.debug( "Setting field {} to value {}", field, value );
+        for( ScenarioStateField field : getScenarioFields( object ) ) {
+            Object value = getValue( field );
+
+            if( value != null ) {
+                try {
+                    field.getField().set( object, value );
+                } catch( IllegalAccessException e ) {
+                    throw new RuntimeException( "Error while updating field " + field.getField(), e );
                 }
-            } catch( IllegalAccessException e ) {
-                throw new RuntimeException( "Error while updating field " + field, e );
+
+                log.debug( "Setting field {} to value {}", field.getField(), value );
+            } else if( field.isRequired() ) {
+                throw new JGivenMissingRequiredScenarioStateException( field.getField() );
             }
         }
     }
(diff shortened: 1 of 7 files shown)


Airblader:
Clearly not ready to merge yet. @janschaefer I assume you prefer a `JGivenWrongUsageException`, right? Do you think the basic approach is OK?

And, lastly, under what version should I make the changelog entry?

janschaefer:
I am currently wondering whether one could also have a required provided. So it would fail if a stage does not provide a certain value when it's finished. What do you think?

jgiven-core/src/main/java/com/tngtech/jgiven/impl/inject/ValueInjector.java
```diff
@@ -90,6 +90,8 @@ public void updateValues( Object object ) {
                 if( value != null ) {
                     field.set( object, value );
                     log.debug( "Setting field {} to value {}", field, value );
+                } else if( isRequired( field ) ) {
+                    throw new RuntimeException( "TODO Better exception?" );
```
janschaefer: Yeah, this should be a different exception. The question is which kind of exception. `JGivenWrongUsageException` is not really fitting, I think, because JGiven is not really used in a wrong way in this case. JGivenPreconditionException/JGivenAssumptionException/JGivenUserException maybe?

Airblader:
It definitely makes sense, in particular because a dependent stage wouldn't have to look at the implementation to see whether the state will always be provided or under which circumstances it will be provided. 

I think we should give this feature more thought in that case. Currently these annotations behave the same and this would make them actually do different things (instead of one of them just not having all the features of the others). 
We either need to accept this "mental breakage" (and should remove the documentation claiming they all behave the same) or make it explicit by naming the attributes differently. Otherwise I'm afraid it becomes a little unintuitive.

jgiven-core/src/main/java/com/tngtech/jgiven/impl/inject/ValueInjector.java
```diff
@@ -90,6 +90,8 @@ public void updateValues( Object object ) {
                 if( value != null ) {
                     field.set( object, value );
                     log.debug( "Setting field {} to value {}", field, value );
+                } else if( isRequired( field ) ) {
```
janschaefer: I wonder whether the code can be optimized so that it does not have to iterate twice over the annotations of the field. Because the getResolution method also does the iteration.

jgiven-junit/src/test/java/com/tngtech/jgiven/junit/ScenarioExecutionTest.java
```diff
@@ -352,4 +352,17 @@ public void abstract_steps_should_appear_in_the_report_model() throws Throwable
 
     }
 
+    static class StageWithMissingState {
+        @ScenarioState( required = true )
```
janschaefer: you could also add a test for `@ExpectedScenarioState`. Maybe also interesting would be a test case where the stage is not the first one

Airblader: Agree, more tests would be good here.

jgiven-core/src/main/java/com/tngtech/jgiven/impl/inject/ValueInjector.java
Airblader: Sure. I'll come up with something. :-)

jgiven-core/src/main/java/com/tngtech/jgiven/impl/inject/ValueInjector.java
Airblader: I think the answer to »can it« is a clear yes. I'm not sure about whether it's worth it, though. Fields generally don't have dozens, hundreds or thousands of annotations, so my first guess is that we're optimizing a non-issue here. Do you have specific concerns about performance for this?

janschaefer:
Yes, I deleted the other comment, because I actually wanted to add that to the review :-). Why not calling the attribute `mandatory`? Then it is universal and can be used in `BeforeScenarioState` and `AfterScenarioState`?

janschaefer:
Did you post your comments twice?

jgiven-junit/src/test/java/com/tngtech/jgiven/junit/ScenarioExecutionTest.java
janschaefer: In particular, there should be a test with multiple steps where the second step sets the value.

jgiven-core/src/main/java/com/tngtech/jgiven/impl/inject/ValueInjector.java
janschaefer: Well, in general I agree, however, JGiven is already doing a lot of reflection stuff, so I like to keep it at a minimum if possible. But I guess it will not be measurable.

Airblader:
> Did you post your comments twice?

No, but I saw it too. But as fast as they appeared they also went away again. Replication issues @ Github, perhaps. :-)

> Why not calling the attribute mandatory?

My feeling was that `required` is more of an idiom here. I couldn't think of any Java library that uses the »mandatory« terminology. Granted, this is highly subjective. :-) Ultimately they're synonyms, anyway.

> Then it is universal and can be used in BeforeScenarioState and AfterScenarioState?

(Assuming you meant to say ExpectedScenarioState / ProvidedScenarioState)

I don't think using `mandatory` solves my concerns. We'd have three annotations which look similar, claim to behave exactly the same but have slightly different semantics for the same interface. This could easily confuse users and, e.g., lead them to believe that ProvidedScenarioState has a meaningless attribute.

By the way, is there a common practice (»Java-wide« or just within JGiven) when to use attributes versus dedicated annotations? I could also imagine having `@Required` (and something like `@Guaranteed` for the other case, perhaps).

jgiven-core/src/main/java/com/tngtech/jgiven/impl/inject/ValueInjector.java
Airblader: The obvious approach here would be to encapsulate annotations so we iterate over them once and »cache« the information we collect. I don't mind adding this if you prefer it.

janschaefer:
I agree, if you have a boolean attribute that you then as well can have another annotation. I am not aware of any common practice here. I think that is a matter of taste. However, in this case, it would make the implementation easier, because we would not have to add the attribute to any of the annotations, so the annotations would all still could behave the same. So we would have something like:

    @ScenarioState @Required
    String name;

    @ScenarioState @Guaranteed
    String message;

The semantics could then be defined for the two new annotations, independently of the three `ScenarioState` annotations/

jgiven-core/src/main/java/com/tngtech/jgiven/impl/inject/ValueInjector.java
janschaefer: Yes that makes sense. Please do so :-)

Airblader:
Fun fact, I tried to do some research on mandatory versus required.

* On Github, 355+ million results for required, but only 15+ million for mandatory.
* On Debian Code Search, 1.5+ million for required and 100k for mandatory.
* According to some linguists, mandatory is primarily used in legal work. Unfortunately no similar statement about required.

---

Yeah, I think dedicated annotations make more sense here for the reasons you mentioned. I'll work all of the feedback in. I'll consider whether I'll add `@Guaranteed` right away or do a separate PR for it. Need to set up Eclipse again first. So far I did this in vim… :-)

janschaefer:
Interesting and mandetory is also harder to spell ;-).
Regarding the IDE: I suggest IntelliJ in this case, because the project generation with ./gradlew idea works out of the box.

janschaefer:
I agree to implement @Guaranteed in another PR.

Airblader:
@janschaefer Actually, can you be more specific on the other test you want to see? Or just find me and we'll talk about it. J

Things left to do:
-[ ] Adapt CHANGELOG (for which version?)
-[ ] Add the missing test case(s)

jgiven-core/src/main/java/com/tngtech/jgiven/impl/inject/ValueInjector.java
```diff
@@ -42,57 +43,65 @@ public void validateFields( Object object ) {
 
         Map<Object, Field> resolvedFields = Maps.newHashMap();
 
-        for( Field field : getScenarioFields( object ) ) {
-            field.setAccessible( true );
-            Resolution resolution = getResolution( field );
+        for( ScenarioStateField field : getScenarioFields( object ) ) {
```
Airblader: Now we go through annotations each time `getScenarioFields` is called, which still happens multiple times even within one method. I think refactoring this would deserve its own PR, though, in particular since this has been existng behavior.

jgiven-core/src/main/java/com/tngtech/jgiven/impl/inject/ValueInjector.java
```diff
@@ -42,57 +43,65 @@ public void validateFields( Object object ) {
...
-            try {
-                Object value = getValue( field );
-                if( value != null ) {
-                    field.set( object, value );
-                    log.debug( "Setting field {} to value {}", field, value );
+        for( ScenarioStateField field : getScenarioFields( object ) ) {
+            Object value = getValue( field );
+
+            if( value != null ) {
+                try {
```
Airblader: Moved these try/catch blocks around so the IllegalAccess block only encapsulates the code that actually can throw this.

janschaefer:
Version would be 0.14.0

Airblader:
@janschaefer We didn't get to continue our conversation today about attribute versus dedicated annotation. :-)

janschaefer:
ok, I thought we agreed on an attribute :-)

janschaefer:
@Airblader: I reverted the use of the ShadowJar plugin. This caused your PR to have merge conflicts now.

Airblader:
@janschaefer Already resolving. I can deal with a few conflicts. :-)

janschaefer:
Wow, you are faster than your Shadow :-D. I already resolved them :-)

Airblader:
@janschaefer Oh, so did I and force-pushed my changes because I did a rebase instead of a merge. :-)

janschaefer:
Ok, that is better :-)

Airblader:
Only that I missed something. :-( I'll update again in a second.

Airblader:
Should be all good now. Just gotta make Eclipse work again… :-)

janschaefer:
One last thing: could you document this feature in the user guide?

PR #283, opened 2016-12-28
Issue 236, Extended Description argument parsing functionality

cirquit:
- added front-end parsing for extended descriptions with arguments
- added front-end tests in utilTest.js
- added example usage in ExtendedDescriptionTest.java
- added example for @As usage with direct references to arguments in AsAnnotationExampleTest.java
- added documentation for @As and @ExtendedDescription in step-usage

missing:
- tests for extended description in html5report


jgiven-examples/src/test/java/com/tngtech/jgiven/examples/description/AsAnnotationExampleTest.java
```diff
@@ -27,6 +30,15 @@ public void scenarios_can_have_an_extended_description() {
             .and().something();
     }
 
+    @Test
+    @As( "Scenario that shows the usage of @At with argument enumeration")
```
Airblader: s/At/As/

jgiven-examples/src/test/java/com/tngtech/jgiven/examples/description/ExtendedDescriptionsTest.java
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    public static class Annotations extends Stage<Annotations> {
+
+        @ExtendedDescription( "This is a boolean value" )
+        public Annotations some_boolean_value ( boolean bool ){
+            return this;
+        }
+
+        @ExtendedDescription( "We can reference the first argument with $$ or $$1 : $" )
```
Airblader: Why do we want this?

cirquit: This was proposed in Isse236, I double escaped the first two occurrences to view the message in the report, but maybe this is misleading

janschaefer:
As discussed offline it would be great to also support substitution by parameter name like `$foo`

janschaefer:
@cirquit I have finished the work on the HTML App extraction. This means, however, that you have to apply your changes to the new code. Sorry about that! On the good side: lodash is the newest version!

jgiven-examples/src/test/java/com/tngtech/jgiven/examples/description/ExtendedDescriptionsTest.java
```diff
@@ -0,0 +1,60 @@
...
+import com.tngtech.jgiven.annotation.Description;
+import com.tngtech.jgiven.annotation.ExtendedDescription;
+import com.tngtech.jgiven.junit.SimpleScenarioTest;
+import com.tngtech.jgiven.tags.Issue;
+import junitparams.JUnitParamsRunner;
+import junitparams.Parameters;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+@RunWith( JUnitParamsRunner.class )
```
janschaefer: We typically use the DataProviderRunner for parameterized tests

jgiven-examples/src/test/java/com/tngtech/jgiven/examples/description/ExtendedDescriptionsTest.java
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    @Test
+    @Parameters ({
+            "false, 0",
+            "true, 1"
+    })
+    public void scenarios_with_multiple_argument_parameters_can_be_shown_via_click_on_table( boolean bool, int i ) {
+        given().some_bool_$_and_int_$_value( bool, i );
+        // TODO screenshot test & implement frontend "click on table"
```
janschaefer: Please move this test to the `jgiven-tests` project, as the examples project is not meant for 'real' tests.

jgiven-html-app/src/js/util.js
```diff
@@ -144,3 +144,65 @@ export function deselectAll (options) {
     option.selected = false;
   });
 }
+
+function getArgumentList(wordList) {
+  function isArgument(word) { return !!word.argumentInfo; }
+  return _.chain(wordList)
```
janschaefer: could you please check whether the use of the `chain` method has increased the artifact size? There seem to be a problem with that function: https://medium.com/making-internets/why-using-chain-is-a-mistake-9bc1f80d51ba#.4s9828vud

jgiven-html-app/src/js/util.js
```diff
@@ -144,3 +144,65 @@ export function deselectAll (options) {
...
+    return result;
+}
+
+function replaceArguments(string, argumentList) {
+    var separator = "$";
+    var result    = [];
+    var placeHolderCount = 0;
+
+    for(var i = 0; i < string.length; ++i) {
+        var c           = string.charAt(i);
```
janschaefer: I am not a big fan of aligning code vertically. I typically automatically format my code, so this hand-crafted formatting will be destroyed the next time I touch the file. I guess we need a coding style guide for the JavaScript code :-)

jgiven-examples/src/test/java/com/tngtech/jgiven/examples/description/AsAnnotationExampleTest.java
```diff
@@ -1,14 +1,17 @@
...
 import com.tngtech.jgiven.annotation.Description;
 import com.tngtech.jgiven.annotation.ExtendedDescription;
 import com.tngtech.jgiven.annotation.IntroWord;
 import com.tngtech.jgiven.junit.SimpleScenarioTest;
+import junitparams.JUnitParamsRunner;
+import junitparams.Parameters;
+import org.junit.Test;
+import org.junit.runner.RunWith;
 
+@RunWith( JUnitParamsRunner.class )
```
janschaefer: Please use the DataProviderRunner instead

jgiven-html-app/src/js/util.js
```diff
@@ -2,6 +2,7 @@
  * Utility functions
  */
 
+let _ = require('lodash');
```
janschaefer: please import 'lodash/core' to reduce the final artifact size. You can also use the ES6 syntax:

     import _ from 'lodash/core'

janschaefer:
I think I found a general problem. Actually, JGiven does not really implement placeholder variables by name. The current behavior is simply that you can give a placeholder an arbitrary name. But the name is only relevant if it is a number. So if you have something like

    @As("Second: $second, First: $first)
        public AsAnnotationStage argument_names_cannot_be_referenced( int $first, int $second ) {
    }

and you call the method like argument_names_cannot_be_referenced( 1, 2 ) will result in

    Second: 1, First: 2 

while this first sounds wrong, however, the basic idea behind that is that you can give your placeholders arbitrary names. In practice, it also makes nearly never sense to change the order of the arguments. However, I still think that this would be useful feature. The problem now is that changing the behavior could break existing code.

For now I suggest that you leave your implementation as it is is, but only document the feature for the @ExtendedDescription annotation and not the @As annotation.

janschaefer:
In a next step I suggest to implement the feature for the `@As` annotation. I think the likelihood that this change will breaking anybodies existing code is very low. As it would mean that someone has used the name of a parameter at a different position and *still* the value of the parameter of the original position was used.

jgiven-examples/src/test/java/com/tngtech/jgiven/examples/description/ExtendedDescriptionsTest.java
cirquit: This will be a example for future use, so I would propose to let this be and add a test after the feature is implemented.

cirquit:
There were some non trivial problems while using webpack with lodash on my side. Based on how I imported the functions, some had a different behaviour and _some_ tests broke.

```
import takeWhile from 'lodash/fp/takeWhile'
import drop from ...
...
```

vs

```
import _ from 'lodash/core'
```

vs

```
import _ from 'lodash'
```

in `util.js`.

The custom import broke almost every function except flow, forEach, curry and filter. The test ran through as if the functions were there, but their behaviour was not as specified (predicates didn't work, `drop` didn't drop enough etc.)...this is very disturbing. 

Your proposed import solution resulted in a test failure, because of the `TypeError undefined is not a constructor (evaluating '_core2.default.functionname...' in utilTest.js)`

I'm trying to get onto why the custom import of functions does not behave the way it should.

And for the new __@As__ functionality, I'm on board with implementing the references by argument names.

cirquit:
Something is kind of strange in the util.js, because the function `deselectAll` seemed to work without ever importing lodash or the `_` object. 

I've also stumbled upon the old code in jgiven-html5-report `src/test/app/*.js`, should this be removed?

cirquit:
I have now a working test where both imported functions from lodash (one with _, and one without) which result in different outputs. The example is the `indexOf` function.

```java
import indexOf from 'lodash/fp';
import _ from 'lodash';

function parseNextArgumentName(string) {
    var stopChars = [" ", ",", ";", ":","\"","%","!","[","]","(",")","-","_"];
    var isNonStopChar = function (c) {
        console.log("char: " + c + ", _.ix: " + _.indexOf(stopChars, c)
                                 + ", ix: " + indexOf(stopChars, c));
        return indexOf(stopChars, c) === -1;
     };
    var res = _.takeWhile(string, isNonStopChar).join("");
    return res;
}
```

The call to `parseNextArgumentName('i, bool : $bool')`results in:

```
LOG: 'char: i, _.ix: -1, ix: -1'
LOG: 'char: ,, _.ix: 1, ix: -1'
LOG: 'char:  , _.ix: 0, ix: -1'
LOG: 'char: b, _.ix: -1, ix: -1'
LOG: 'char: o, _.ix: -1, ix: -1'
LOG: 'char: o, _.ix: -1, ix: -1'
LOG: 'char: l, _.ix: -1, ix: -1'
LOG: 'char:  , _.ix: 0, ix: -1'
LOG: 'char: :, _.ix: 3, ix: -1'
LOG: 'char:  , _.ix: 0, ix: -1'
LOG: 'char: $, _.ix: -1, ix: -1'
LOG: 'char: b, _.ix: -1, ix: -1'
LOG: 'char: o, _.ix: -1, ix: -1'
LOG: 'char: o, _.ix: -1, ix: -1'
LOG: 'char: l, _.ix: -1, ix: -1'
```

The manually imported`indexOf`function can't find the comma in the array, but the automatically imported one can...

janschaefer:
The point is that you import from the 'fp' module of lodash. The functions of this module are different. I would suggest to not use this for now to avoid these kinds of problems.

janschaefer:
Regarding 

> I've also stumbled upon the old code in jgiven-html5-report src/test/app/*.js, should this be removed?

I have removed them on master now.

janschaefer:
There is another minor issue regarding consistency between the `@As` implementation and your implementation, regarding spaces after placeholders. In JGiven it is currently not possible to *not* have a space after a placeholder. This means that something like `@As('foo, $1, baz')` where `$1` is replaced with `bar` will result in `foo, bar , baz`. JGiven will currently always create a space after a placeholder. I also suggest that for the `@ExtendedDescription` case you should implement it as expected, namely without the space and we try to fix it later for the `@As` annotation.

cirquit:
I fixed the imports now, this is still a very strange behaviour of the `flow`function, because it is implemented the same way as in `lodash/fp` regarding the argument order of the inner functions. One needs to use `partial` almost for every library function that takes the arguments in the following order `([collection], function)`.

cirquit:
I think I found another error, since the merge to the new jgiven-html-app the tests in HtmlReport don't work anymore. I tested it on my issue branch and [the branch where the new html-app was introduced](https://github.com/TNG/JGiven/commit/cfb5ef895055e3c13892eb2bbf0d1d3a3b4baa9c).

The problem seems to be [this](https://docs.angularjs.org/error/$injector/modulerr?p0=ng&p1=TypeError:%20%27undefined%27%20is%20not%20an%20object%20(evaluating%20%27Function.prototype.bind.apply%27)%0A%20%20%20%20at%20instantiate%20(file:%2F%2F%2Ftmp%2Fjunit4194905036476466323%2FtargetReportDir%2Fapp.bundle.js:18893)%0A%20%20%20%20at%20provider%20(file:%2F%2F%2Ftmp%2Fjunit4194905036476466323%2FtargetReportDir%2Fapp.bundle.js:18707)%0A%20%20%20%20at%20file:%2F%2F%2Ftmp%2Fjunit4194905036476466323%2FtargetReportDir%2Fapp.bundle.js:14525%0A%20%20%20%20at%20forEach%20(file:%2F%2F%2Ftmp%2Fjunit4194905036476466323%2FtargetReportDir%2Fapp.bundle.js:14495)%0A%20%20%20%20at%20file:%2F%2F%2Ftmp%2Fjunit4194905036476466323%2FtargetReportDir%2Fapp.bundle.js:18697%0A%20%20%20%20at%20ngModule%20(file:%2F%2F%2Ftmp%2Fjunit4194905036476466323%2FtargetReportDir%2Fapp.bundle.js:16730)%0A%20%20%20%20at%20invoke%20(file:%2F%2F%2Ftmp%2Fjunit4194905036476466323%2FtargetReportDir%2Fapp.bundle.js:18878)%0A%20%20%20%20at%20runInvokeQueue%20(file:%2F%2F%2Ftmp%2Fjunit4194905036476466323%2FtargetReportDir%2Fapp.bundle.js:18771)%0A%20%20%20%20at%20file:%2F%2F%2Ftmp%2Fjunit4194905036476466323%2FtargetReportDir%2Fapp.bundle.js:18780%0A%20%20%20%20at%20forEach%20(file:%2F%2F%2Ftmp%2Fjunit4194905036476466323%2FtargetReportDir%2Fapp.bundle.js:14481)%0A%20%20%20%20at%20loadModules%20(file:%2F%2F%2Ftmp%2Fjunit4194905036476466323%2FtargetReportDir%2Fapp.bundle.js:18803)%0A%20%20%20%20at%20createInjector%20(file:%2F%2F%2Ftmp%2Fjunit4194905036476466323%2FtargetReportDir%2Fapp.bundle.js:18683)%0A%20%20%20%20at%20file:%2F%2F%2Ftmp%2Fjunit4194905036476466323%2FtargetReportDir%2Fapp.bundle.js:15918%0A%20%20%20%20at%20bootstrap%20(file:%2F%2F%2Ftmp%2Fjunit4194905036476466323%2FtargetReportDir%2Fapp.bundle.js:15939)%0A%20%20%20%20at%20angularInit%20(file:%2F%2F%2Ftmp%2Fjunit4194905036476466323%2FtargetReportDir%2Fapp.bundle.js:15824)%0A%20%20%20%20at%20file:%2F%2F%2Ftmp%2Fjunit4194905036476466323%2FtargetReportDir%2Fapp.bundle.js:45923%0A%20%20%20%20at%20file:%2F%2F%2Ftmp%2Fjunit4194905036476466323%2FtargetReportDir%2Fapp.bundle.js:7496%0A%20%20%20%20at%20file:%2F%2F%2Ftmp%2Fjunit4194905036476466323%2FtargetReportDir%2Fapp.bundle.js:7564), and with my little knowledge of angular it might be that this call to `filter`:
```java
jgivenReportApp.filter('encodeUri', function ($window) {
    return $window.encodeURIComponent;
})
```
in [reportApp.js](https://github.com/TNG/JGiven/blob/master/jgiven-html-app/src/js/reportApp.js) is the culprit.

With the previous minor version of `angular-1.4.8` instead of `1.5.8` only one of the tests fails, namely [this one, Line 96](https://github.com/TNG/JGiven/blob/master/jgiven-tests/src/test/java/com/tngtech/jgiven/report/html5/Html5AppTest.java) . The tested functionality works if tested manually, but this test throws an error because of the WebDriver, that can't find any tags. 

The error is thrown [here](https://github.com/TNG/JGiven/blob/master/jgiven-tests/src/test/java/com/tngtech/jgiven/report/html5/Html5AppStage.java) in the findTagWithName method. After changing `By.xpath` to a simplified `By.linkText` it found the tag, but the `click()` didn't work and the title of the scenario was still `All Scenarios` instead of the `[TagName]`.

This may be still the problem of angular.

cirquit:
Looks like this is only a problem on my ubuntu box:
````
Ubuntu 15.10 wiliy
npm 2.14.12
yarn 0.18.1
oracle javac 1.7.0_80
````

The mac, as well as travis don't seem to have any issues.

**Comment by janschaefer** [comment]
Ok. Then ignore the issues on your machine for now :-)

File: jgiven-tests/src/test/java/com/tngtech/jgiven/report/html5/Html5AppTest.java [hunk 9]
```diff
@@ -238,27 +239,27 @@ public void newlines_are_detected_in_formatted_values_and_shown_as_multiline_tex
...
+                + " and this the third named one: 3";
 
         given().a_report_model()
                 .and().step_$_of_scenario_$_has_extended_description_with_arguments( 1, 1, description, argumentMap );
         jsonReports
                 .and().the_report_exist_as_JSON_file();
         whenReport.when().the_HTML_Report_Generator_is_executed();
-        when().the_page_of_scenario_$_is_opened( 1 );
+        when().the_page_of_scenario_$_is_opened( 1 )
+              .and().show_angular_foundation_tooltip_with_$_ms_delay_for_element_$_with_attribute_$_and_value_$(251, "span", "class", "has-tip" );
```
janschaefer: You should introduce another step here that hides the internal details of the implementation. Always think from the perspective of a non-technical person, when writing the scenarios.
For example: `the_tooltip_of_extended_description_is_shown()`. That method can then call your internal implementation.

jgiven-tests/src/test/java/com/tngtech/jgiven/report/html5/WhenHtml5App.java
```diff
@@ -65,4 +70,14 @@ public void takeScreenshotAfterStage() {
         takeScreenshot();
 
     }
+
+    public SELF show_angular_foundation_tooltip_with_$_ms_delay_for_element_$_with_attribute_$_and_value_$( long ms, String element, String attribute, String value){
+        Actions builder = new Actions(webDriver);
+        WebElement webElement = webDriver.findElements( By.xpath( "//" + element + "[@"+ attribute +"='" + value + "']" )).get( 0 );
+        builder.moveToElement( webElement ).clickAndHold().build().perform();
+        WebDriverWait wait = new WebDriverWait( webDriver, ms );
+        wait.until( ExpectedConditions.presenceOfElementLocated( By.xpath( "//span[@is-open]" )));
```
janschaefer: ok, that is nice!

jgiven-tests/src/test/java/com/tngtech/jgiven/report/html5/Html5AppTest.java
cirquit: I thought of it as the person looking at the test, and if it didn't work because of the newly introduced framework that doesn't use angular-tooltips, I would want to know this dependency explicitly.

PR #286, opened 2016-12-29
Generate language-dependent classes (fixes #285)

### LINKED ISSUE #285: Implement a mechanism to generate language-specific classes from language files
Currently, language-specific classes have to be coded by hand, which introduces a lot of unnecessary duplication and makes translations more difficult. Instead of writing the classes by hand one could generate the classes from language files.

janschaefer:


Airblader:
That was quick. :-)

janschaefer:
And fun :-)

PR #287, opened 2016-12-31
Extract HTML App into separate project + Update to modern build tools

### LINKED ISSUE #238: Extract HTML 5 Report into separate project
In order to make it easier to develop the HTML report independently of JGiven, it would be useful to put the HTML report into its own project with a separate release cycle. JGiven should still provide a Java artifact with a version matching that of JGiven.

As the HTML 5 report is a pure Javascript project, it would make sense to publish it to NPM.

This is in particular important to easier support different frontends, like https://github.com/jsGiven/jsGiven. Also see https://github.com/jsGiven/jsGiven/issues/53.

janschaefer:
This PR provides the following:

* The JavaScript+HTML part of the HTML5 Reporter is moved to a new project `jgiven-html-app`
  - This allows, in a next step, to even move it to a complete new repository as there are no dependencies on jgiven-core, and to publish the app to NPM, for example.
* The build is completely based on webpack+yarn now (no bower modules anymore)
* Updated dependencies as long as this was possible without major effort (AngularJS, for example, could only be updated to 1.5.8 without major issues

This PR fixes #238 

Still open:
 
- [x] Port #242 
- [x] Version is not correctly substituted


Airblader:
This is a really cool change. :-)

janschaefer:
Thanks :-)

Issue #252, opened 2016-11-17
Use formatters on field level in @Table POJOs


nikowitt:
Hi Jan,

I currently have to implement test scenarios that heavily rely on the @Table annotation.

```

	private static final Code RATE_11_TO_20 = Code.of(ScaffoldCode.class, 11, 20, 16, 8, 0.0171);
	private static final Code RATE_31_PLUS = Code.of(ScaffoldCode.class, 31, 1000, 13, 6.5, 0.0171);

	@Test
	public void first_scenario() {
		given().codes_with_rates(RATE_11_TO_20, RATE_31_PLUS);
		(...)
	{

(...)

	public CostCalculationStage codes_with_rates(@Table Code... rates) {
		(...)
	}

```

where Code is
```
	static class Code {

		private final Class<? extends BaseCode<?>> clazz;
		private final int minSize;
		private final int maxSize;
		private final double erection;
		private final double dismantling;
		private final double rent;

		private Code(Class<? extends BaseCode<?>> clazz, int minSize, int maxSize, double erection, double dismantling, double rent) {
			this.clazz = clazz;
			this.erection = erection;
			this.dismantling = dismantling;
			this.rent = rent;
			this.minSize = minSize;
			this.maxSize = maxSize;

		}

		public static Code of(Class<? extends BaseCode<?>> clazz, int minSize, int maxSize, double erection, double dismantling, double rent) {
			return new Code(clazz, minSize, maxSize, erection, dismantling, rent);
		}

		public double getErection() {
			return erection;
		}

		public double getDismantling() {
			return dismantling;
		}

		public double getRent() {
			return rent;
		}

		public int getMinSize() {
			return minSize;
		}

		public int getMaxSize() {
			return maxSize;
		}

		public Class<? extends BaseCode<?>> getClazz() {
			return clazz;
		}

	}
```

creates the output:

```
   Given codes with rates

     | clazz                                       | minSize | maxSize | erection | dismantling |   rent |
     +---------------------------------------------+---------+---------+----------+-------------+--------+
     | class com.sobis.sms.model.code.ScaffoldCode |      11 |      20 |     16.0 |         8.0 | 0.0171 |
     | class com.sobis.sms.model.code.ScaffoldCode |      31 |    1000 |     13.0 |         6.5 | 0.0171 |
```

It would be really cool if in this case, formatter annotations could also be recognized on field level, e.g.

```
static class Code {
(...)
@SimpleClassName
[...]

Issue #122, opened 2015-09-14
Combine @Table with another formatter?


nikowitt:
Hi again,

I'm currently playing again with the @Table annotation:

```
		public SupplierTestScenario quotations_of_the_same_company_are_awarded_on(
				@Table(numberedRows = true) DateTime... awardDates) throws Exception {
			for (int i = 0; i < awardDates.length; i++) {
				quotation_$_of_the_same_company_is_awarded_on(i + 1, awardDates[i]);
			}
			return self();
		}
```

returns the result:

```
    When quotations of the same company are awarded on

     | # |       iMillis | iChronology                  |
     +---+---------------+------------------------------+
     | 1 | 1356994800000 | ISOChronology[Europe/Berlin] |
     | 2 | 1420066800000 | ISOChronology[Europe/Berlin] |
     | 3 | 1388530800000 | ISOChronology[Europe/Berlin] |
```
By default, all POJO fields are rendered (unless modified with includeFields/excludeFields). The usage of fields basically works fine for POJOs where I have full control of or where I can find fields fields that return a well-formatted value.

In my case, DateTime consists of the fields iMillis or iChronology which are both not very readable. Therefore, my idea is to be able to supply an additional formatter.

If this makes sense for you, there are IMHO two ways:
a) The simple approach which probably already works without too much effort results in a one row/column (depending on the formatting) table + a optional numbering row/column. This would (at least for me... currently :) ) meet my requirement as I then would define my method like this
```
		public SupplierTestScenario quotations_of_the_same_company_are_awarded_on(
				@Table(numberedRows = true) @DateOnly DateTime... awardDates) throws Exception {
```
and receive an output like
```
    | # |       dateOnly | 
     +---+---------------+
     | 1 | 2013-01-01|
     | 2 | 2015-01-01 |
     | 3 | 2014-01-01 | 
```

Another flavour of this idea is to not be able to invoke the toString() method of a POJO, although the formatter idea is more flexible and powerful.

b) The other approach provided a way of post-formatting more than one value. There are several ideas of doing this (Formatter has to return an array? Be able to supply more than one formatter?).

Perhaps this request also makes sense to be included in JGiven.

Best regards,
Niko

janschaefer:
[...]

PR #297, opened 2017-01-18
feature suggestion to format fields inside a table

dgrandemange:
Hi,

Lately, i needed to provide a way to format fields inside a table.
Don't know if my implementation is the best way to do it, but as it fulfills my own needs, it may interest other users.
I've seen that there was an issue already talking about this need, but the provided solution doesn't well fit my needs at it requires the POJO fields to be Format annotated. Well, it's ok when the source of this POJO is under control. Besides i don't like much the idea of intermingling presentation aspect into domain POJO. 

Anyway, you may get a demo by running the jgiven-examples tests.

Damien


janschaefer:
Thanks for the PR! I will have a look at it as soon as possible.

janschaefer:
Related to #252

janschaefer:
Btw. there is already a way to format fields inside tables. Did you have a look at https://github.com/TNG/JGiven/issues/122?

dgrandemange:
Yes, i indeed read about formatter() and rowFormatter() (without experimenting it though) but truth is my motivations here were clearly : 
- to re-use something that i know how to use (the @Format annotation), 
- re-benefit my investment (reuse already coded specific ArgumentFormatter implementations), 
- make the all thing simple (annotation declaration) and now re-usable (see previous commit [0a82a57](https://github.com/TNG/JGiven/pull/297/commits/0a82a576fe3b7f9befa84c53c9ba0c60be7aacc8))

dgrandemange:
More precisely, i recall i first began by trying to specify my own implementation of `RowFormatter`. I've tried to extend the `FieldBasedRowFormatter` class, and override the `List<String> formatRow(Object object)` method for my particular fields formatting needs. Still, to be able to do this, I was needing access to the `FieldBasedRowFormatter` class members `fields `and `nonNullColumns` that unfortunately were private.
Given that, I eventually ended up putting my fields formatting logic directly into the `FieldBasedRowFormatter`.

jgiven-core/src/main/java/com/tngtech/jgiven/format/table/FieldBasedRowFormatter.java
```diff
@@ -47,14 +56,63 @@ public FieldBasedRowFormatter( Class<?> type, String parameterName, Table tableA
...
+                    .isAnnotationPresent(TableFieldsFormats.class)) {
+                tffs = (TableFieldsFormats) annotationType
+                        .getAnnotation(TableFieldsFormats.class);
+            }
+
+            if (tffs != null) {
+                TableFieldFormat[] tffTab = tffs.value();
+                for (TableFieldFormat tff : tffTab) {
+                    String fieldName = tff.value();
+                    Format format = tff.format();
```
janschaefer: a little disadvantage here is that you cannot have custom annotations. This could make this a bit cumbersome to use. It would be great if all ways of formatting could be supported.

dgrandemange:
Hi, thank you for the review.

> I would suggest to integrate this functionality into the @Table annotation. I see no benefit of an extra annotation. In particular, as it has to be always together with the @Table annotation.

I agree it makes sense to have this functionality into the `@Table` annotation. 
Still, i can see a case where keeping this `@TableFieldsFormats` annotation (kind of wrapper of `@TableFieldFormat` set for a specific bean) could also makes sense outside of a `@Table` formatting context, for instance in a simple @Format context, where a new `ArgumentFormatter<?>` implementation could use this annotation in order to render a single bean (without needing to redefine all its related `@TableFieldFormat` once again). Got to think about this.

> We should think how we can support custom format annotation with your solution

By _custom format annotation_, do you mean an annotation which itself is annotated with the `@Format` annotation ? such as the `@Quoted` sample annotation mentioned in the documentation guide ? 
Yes, i've 2-3 things and i think it's possible, as the #252 support.

I'll try to provide something next week.

dgrandemange:
Hi,

Every requested changes have been addressed.
Documentation has also be consequently completed (see "Parameterized Steps" section).
Enjoy.

Damien

janschaefer:
Amazing! Thank you very much! I am currently quite busy so it might take some time until I can do a review

janschaefer:
Thank you very much for the PR and sorry for the late response. I have merged the PR on the command line as I did some minor renamings.

dgrandemange:
Thanks for the info.
I've been playing around a month ago with some local build of my PR, and somehow, some pattern emerges. So i share it now.

Let's say you have a POJO of type `IssuingBank` :

```
    public class IssuingBank {
        private String bankCode;
        private String name;
        private Integer maxTokensPerCard;

        // ...
    }
```

and you have defined the following annotation to define its fields format (quite straight here for the sake of the understanding)

```
    @NamedFormatSet({ 
            @NamedFormat(name = "bankCode"), 
            @NamedFormat(name = "name"),
            @NamedFormat(name = "maxTokensPerCard") 
    })
    @Retention(RetentionPolicy.RUNTIME)
    public @interface IssuingBankFieldsFormatSet {
    }
```

So, this format is now usable to render a single (or a table) of `IssuingBank` instances in jGiven reports.
    
Now, let's say you also have another type of POJO, `BinRange`, whose has one complex field of the type described above :
    
```
    public class BinRange {
        private String lower;
        private String upper;
        private IssuingBank issuingBank;

        // ...
    }
```

When rendering such `BinRange` pojo, having the issuingBank field to be rendered only by its code (along its name maybe) would be quite enough for the report reader to be able to do cross-checks. 
Idea here is to define some reduced/lightweight version of our previous `IssuingBankFieldsFormatSet`. We can achieve this by defining a new format like this :

```
    @POJOFormat(
            fieldsFormatSetAnnotation = IssuingBankFieldsFormatSet.class, 
            includeFields = {"bankCode", "name" }, 
            prefixWithFieldName = true)
    @Retention(RetentionPolicy.RUNTIME)
    public @interface IssuingBankReducedFormat {
    }
```

(Please note the use of the `includeFields` attribute to select the relevant fields here.)

Finally, we can define `BinRange` POJO format set by referencing this `IssuingBankReducedFormat`, like this :

```
    @NamedFormatSet({ 
            @NamedFormat(name = "lower"), 
            @NamedFormat(name = "upper"),
            @NamedFormat(name = "issuingBank", customFormatAnnotation = IssuingBankReducedFormat.class) })
    @Retention(RetentionPolicy.RUNTIME)
    public @interface BinRangeFieldsFormatSet {
    }
```

Eventually, you end up having one long format version, and one short format version of your POJOs. 
But cool thing is  : the short version is somehow an inheritance from the long version. You don't have to repeat the fields formats definitions but only tell wich ones you want to select.

Issue #275, opened 2016-12-25
HTML Report: Have the option to show an image attachment as a popup


janschaefer:
Currently when clicking to an attachment icon a new browser page is opened. It would be more convenient, in particular for images, if a popup would be shown that shows the image. It might be useful to have this configurable.

cirquit:
Should this be a popup that is embedded in the html and is activated via a click? The other option would be an image that pops up if you hover over the thumbnail, and is gone after you move the mouse away

janschaefer:
It should not open when hovering, but when clicking on the thumbnail. Then an overlay should be shown with the image, so that the attachment is not opened in another tab, for example.

janschaefer:
See: http://foundation.zurb.com/sites/docs/v/5.5.3/components/reveal.html

cirquit:
It seems like none of these snippets want to work in the report. The buttons exist and are clickable, but no modal is shown. The modal doesn't change the value `visibility: hidden` and `display:none` after a click.

If these are changed manually the dialog is visible but not closable.

janschaefer:
Mmmh, maybe the Foundation Javascript is not included.

cirquit:
I included the foundation javascript and implemented the popup. There are still some minor errors:

- How can I include the foundation.min.js to every build? (I've done it manually until now)
- After a reload the `$(document).foundation()` at the end of the `index.html` does not seem to work. I have to run it in the console to take action
- The group-by bar suddenly disappears if I click on any image

Here is a [webm](https://webmshare.com/KE1YM) of the last problem.
My fork of the current version: https://github.com/cirquit/JGiven/tree/issue275-image-preview-on-click

janschaefer:
The point should be easily handled by just adding an import statement to the 'entry.js' file. The 2nd and 3rd point, I don't know.

janschaefer:
Closing the issue here as it will be handled in https://github.com/jgiven/jgiven-html-app/issues/4

PR #299, opened 2017-01-25
implemented thumbnail preview for every picture attachment

cirquit:
* added javascript as well as selenium tests for this functionality
* currently hardcoded scaling and saving the thumbs in the attachment directory

Looks like this:
screen shot 2017-01-25 at 13 28 21


jgiven-html-app/src/js/util.js
```diff
@@ -225,4 +225,17 @@ export function replaceArguments(string, enumArray, nameArray) {
         }
     }
     return result.join("");
+}
+
+export function insertThumbnailPath(path) {
```
janschaefer: I would rather call this function `getThumbnailPath`. Maybe you could also add a short documentation.

jgiven-html-app/src/test/utilTest.js
```diff
@@ -271,4 +271,35 @@ describe("Util", function () {
...
+
+  describe("substituteThumbnailPath", function() {
+      it("works for paths with no numbers", function() {
+           var path0 = "/simple/data/path.png";
+           expect(insertThumbnailPath(path0)).toEqual("/simple/data/path-thumb.png")
+      });
+
+      it("works for paths with single number", function() {
+           var path0 = "/simple/data/path0.png";
+           expect(insertThumbnailPath(path0)).toEqual("/simple/data/path-thumb0.png");
```
janschaefer: Can we have that simpler and just have `path0-thumb.png`?

jgiven-html5-report/src/main/java/com/tngtech/jgiven/report/html5/Html5AttachmentGenerator.java
```diff
@@ -105,6 +112,9 @@ private File writeFile( AttachmentModel attachment, MediaType mediaType ) {
         File targetFile = getTargetFile( attachment.getFileName(), extension );
         try {
             if( attachment.isBinary() ) {
```
janschaefer: I think that is not enough. The attachment could be binary, but no image (e.g. a ZIP file).

jgiven-html5-report/src/main/java/com/tngtech/jgiven/report/html5/Html5AttachmentGenerator.java
```diff
@@ -115,4 +125,41 @@ private File writeFile( AttachmentModel attachment, MediaType mediaType ) {
         return targetFile;
     }
 
+    private String compressToThumbnail( String base64content, String extension ) {
+        byte[] imageBytes = javax.xml.bind.DatatypeConverter.parseBase64Binary(base64content);
```
janschaefer: I would extract a method here that takes a BufferedImage and produces one and do the Base64 conversion outside of that.

jgiven-html5-report/src/main/java/com/tngtech/jgiven/report/html5/Html5AttachmentGenerator.java
```diff
@@ -105,6 +112,9 @@ private File writeFile( AttachmentModel attachment, MediaType mediaType ) {
         File targetFile = getTargetFile( attachment.getFileName(), extension );
         try {
             if( attachment.isBinary() ) {
+                File thumbFile = getTargetFile( "attachment-thumb", extension );
+                String thumbnail = compressToThumbnail( attachment.getValue(), extension );
+                Files.write( parseBase64Binary( thumbnail ), thumbFile );
```
janschaefer: That does not make sense to me, you first encode the thumbnail to Base64, to immediately decode it afterwards again?

jgiven-html-app/src/test/utilTest.js
cirquit: this would mean that the `getTargetFile` method from `Html5AttachmentGenerator`would have to be duplicated for thumbnails, mainly because of the enumeration strategy via multisets

cirquit:
Addressed every issue and modified a test because of the now working automatic conversion to thumbnails of images.

jgiven-html-app/src/test/utilTest.js
cirquit: I'm not sure if my solution to this is nice. I tried to minimize duplicate code, but the explicit using filehandles as an argument is maybe not the good practice.

cirquit:
This solves #275

janschaefer:
One last thing: could you add an entry to the changelog, pls?

cirquit:
As discussed offline, this needs to be a configurable option. This will need a little bit of refactoring:

- [ ] destill the global config for every possible ReportGenerator
- [ ] create an abstract config parser for cmd arguments and give a specialized one to every report generator so we don't have a `--customjs=` flag for the AsciiReport
- [ ] implement a `--without-thumbs` flag for the html Report
- [ ] write this information in the meta-data of the html report and update the index.html to show the clip instead of the thumbnail

cirquit:
I updated the interface for the reports, would you mind taking a look at it?

jgiven-core/src/main/java/com/tngtech/jgiven/report/AbstractReport.java
```diff
@@ -0,0 +1,184 @@
...
+ *   <li> --title= </li>
+ *   <li> --exclude-empty-scenarios=&lt;boolean&gt; </li>
+ *   <li> --help / -h </li>
+ * </ul>
+ * Everything except the {@link ReportGenerator.Format} has a default value.
+ *
+ * The functionality is piped together for an easier and extendable interface to create a custom report.
+ *
+ */
+public abstract class AbstractReport {
```
janschaefer: Please name it AbstractReportGenerator

jgiven-core/src/main/java/com/tngtech/jgiven/report/AbstractReport.java
```diff
@@ -0,0 +1,184 @@
...
+        return excludeEmptyScenarios;
+    }
+
+    private void setExcludeEmptyScenarios( Boolean excludeEmptyScenarios ) {
+        this.excludeEmptyScenarios = excludeEmptyScenarios;
+    }
+
+    public void addFlags( String... flags ) {
+        for( String flag : flags ) {
+            this.flags.add( flag );
```
janschaefer: there should be an addAll Method that already does the loop for you

jgiven-core/src/main/java/com/tngtech/jgiven/report/AbstractReport.java
```diff
@@ -0,0 +1,184 @@
...
+        }
+
+        setExcludeEmptyScenarios( Boolean.parseBoolean( getOrDefault( flagMap, "--exclude-empty-scenarios=", "false" ) ) );
+        parseFlags( flagMap );
+    }
+
+    public void printUsageAndExit() {
+        System.err
+                .println( "Options: \n"
+                        + "  --format=<format>             the format of the report. Either html or text (required) \n"
```
janschaefer: It could be overkill, but I wonder whether you could not generalize your code further by introducing an CommandLineOption class that contains all information about a command line option, like description, deprecation and so on. Then you code will become more general in the end.

jgiven-gradle-plugin/src/main/java/com/tngtech/jgiven/gradle/internal/AbstractJGivenReportImpl.java
```diff
@@ -18,13 +18,13 @@ public AbstractJGivenReportImpl( String name, Task task, String relativeEntryPat
...
 
     @Override public void configure( ReportGenerator generator ) {
-        generator.setTargetDirectory( getDestination() );
-        generator.setFormat( getFormat() );
-        generator.getConfig().setCustomCssFile( getCustomCssFile() );
-        generator.getConfig().setCustomJsFile( getCustomJsFile() );
-        generator.getConfig().setExcludeEmptyScenarios( isExcludeEmptyScenarios() );
-        if (getTitle() != null) {
-            generator.getConfig().setTitle( getTitle() );
+        generator.addFlag( "--targetDir=" + getDestination() );
```
janschaefer: I don't really like introducing String constants here. Why not keeping the old methods? Maybe it would be useful to separate the configuration option logic into a class that is separated from the generator?

cirquit: The decision to use string constants benefits the extensibility of the ReportGenerator. It only knows the format and calls the respective report based on that, but it should not know the implementation details of that report. 

The ReportGenerator is mainly a command line interface + the added interface to provide it with flags when instantiated as an object which has a single end point and doesn't have to be reimplemented `generate(List<String> args`.

The problem lies in the custom flags that every report may have - how does one access these custom configurations from the ReportGenerator that sees only an AbstractReport? 

A compromise would be to set all already possibly known flag options like `sourceDir/targetDir` via a custom function, but every custom flag would still be passed as string, but that would be a weird interface to use.

I thought that the unified access point should be the command line flags, as this is the main usage of the reports.

jgiven-core/src/main/java/com/tngtech/jgiven/report/AbstractReport.java
cirquit: I will follow this train of thought.

jgiven-gradle-plugin/src/main/java/com/tngtech/jgiven/gradle/internal/AbstractJGivenReportImpl.java
janschaefer: The problem I see is that you have more or less introduced the command-line frontend into the backend code. This is typically a bad idea. It would be nicer to have clean separation of the frontend (parsing) and the backend. So that the backend does not have to know anything about how command line arguments are parsed/have to look like etc. So my suggestion would be the following:
- Introduce a class for a command line option, something like:

      class CommandLineOption {
           String longName; // mandatory, can be used as key later
           String shortName;
           String description;
           boolean hasValue;
           boolean optional;
           Object defaultValue;
      }

- Have a generic method in the AbstractReportGenerator to specify additional CommandLineOption 
- Implement a CommandLineOptionParser that takes a list of CommandLineOption and an array of Strings and returns a map from String to Object, where String is the longName of the argument and Object the parsed value.

- Have a generic method (in the command line parser) to get the value for an CommandLineOption like:
  
      public Object getValue( CommandLineOption option) {
            return parsedOptions.get(option.getLongName());
      }

If we are at it already, we could even go a step further and generalize this to completely abstract away where the configuration options come from. Something similar to what the TNG config-builder does: https://github.com/TNG/config-builder (but simpler, not using annotations etc). Because in the end, we also want to provide a way to configure the cssFile, for example, using properties. So we may need a general configuration option class like:

      class ConfigurationOption {
             String longName;
             String shortName;
             String propertyString;
             String envString;
             boolean optional;
             boolean hasValue;
             String defaultValue;
             ...
      }

You should also provide a Builder class for that so that it easier to create instances of that class. And instead of only getting these options by using the command line arguments, we would also look for properties and environment variables.

In the end however, there should also be a class like `ReportConfig` and `HtmlReportConfig` that only contain fields with the final configuration values:

      class ReportConfig {
             File targetDirectory;
             ...
      }

      class HtmlReportConfig extends ReportConfig {
             File cssFile;
      }

This is can then be used by the backend code.

cirquit:
Refactored the Report creation with ConfigOptions. Some things are still kind of not optimal:

- I had to let the AbstractConfig in AbstractConfigGenerator be public, because the gradle-plugin has two-step configuration (`JGivenReportTask#generate` and `JGivenReport#createGenerator` (previous version was named `configure`)
- The encoding of the convertion function is done via an interface `com.tngtech.jgiven.report.config.converter.StringConverter` since we are using Java 7. There may be a better option to do this
- the tests in `com.tngtech.jgiven.report.WhenReportGenerator` have to have the three different configurations so the corresponding reports get the correct instantion of the `AbstractReportConfig`. I don't know if my solution to downcast the `AbstractReportConfig` is the right way to create an interface for passing the configuration from the plugins or tests
- the `ConfigOption` is prepared for the property / environment parsing as well as the `ConfigOptionParser` but is not implemented yet

cirquit:
Besides adding a stronger coupling for the ConfigOptions in the AbstractConfig, I added also a suggestion algorithm if the flags have typos.
suggestions

This is done automatically for any future flags as well.

Issue #282, opened 2016-12-27
 @As and @CaseDescription annotation use different placeholder enumerations


cirquit:
If I want to adress the first argument in the `@As ( "First argument: $1, Second Argument: $2" )`, it starts with the index 1. In **@CaseDescription** it starts with index 0.

I would propose to set the **@As** index to start from 0, because usually programmer will write these tests.

Need to add at least two examples for simple placeholder and enumerated ones to docs and jgiven-examples.

`@As ( "First argument: $, second argument $")`
`@As ( "Second argument: $1, first argument $0")`

`@CaseDescription ( "First argument: $, second argument $")`
`@CaseDescription ( "Second argument: $1, first argument $0")`

CaseDescription is also not documented yet.

Minimal example for **@CaseDescription** is found [here](https://github.com/TNG/JGiven/blob/master/jgiven-junit/src/test/java/com/tngtech/jgiven/junit/JUnitParamsTest.java).

Minimal example for **@As** Description:
``` java

import junitparams.JUnitParamsRunner;
import junitparams.Parameters;
import org.junit.Test;

import com.tngtech.jgiven.Stage;
import com.tngtech.jgiven.annotation.As;
import com.tngtech.jgiven.annotation.Description;
import com.tngtech.jgiven.annotation.IntroWord;
import com.tngtech.jgiven.junit.SimpleScenarioTest;
import org.junit.runner.RunWith;

RunWith( JUnitParamsRunner.class )
@Description( "Demonstrates the usage of the @As annotation" )
public class AsAnnotationExampleTest extends SimpleScenarioTest<AsAnnotationExampleTest.AsAnnotationStage> {

    @Test
    @As( "Scenario that shows the usage of @At with argument enumeration")
    @Parameters({
            "false, 0"
    })
    public void steps_can_use_at_annotation_to_reference_arguments_by_enumeration(boolean bool, int i) {
        given().some_boolean_and_int_value( bool, i );
    }

    public static class AsAnnotationStage extends Stage<AsAnnotationStage> {

        @As( "Referencing first argument : $1, second argument : $2 ")
        public AsAnnotationStage some_boolean_and_int_value( boolean bool, int i) {
            return this;
        }
    }
}
```

Airblader:
This is unfortunate, but I'd prefer starting at 1. I expect a programmer to know both, but not somebody else. It also feels more natural.

janschaefer:
[...]

PR #301, opened 2017-01-31
Issue282 - Deprecated Case Annotation

cirquit:
* deprecated the case annotation for `OrdinalCaseAnnotataion` as described in #282 
* added documentation
* because the documentation used asciidoc examples, I also implemented asciidoc parsing for case descriptions


cirquit:
This code produces this following ascii-document.
ordinal-desc-coffe
ordinal-desc-asciidoc

janschaefer:
Ok. Looks good so far, the only thing: I don't like the name OrdinalCaseDescription :-). You immediately wonder what is a normal CaseDescription which is deprecated. Maybe we find another name that is as general as CaseDescription?

cirquit:
From the top of my head:

- `TableDescription`
- `RowDescription`
- `ItemDescription`
- `CaseDataDescription`
- `UnitDescription`
- `ElementDescription`

cirquit:
As discussed offline, before merging I want to remove all ambiguity from CaseDescription AND As Annotation, because it does not allow mixing of placeholders like the ExtendedDescription.

- [X] Add `$argumentName` parsing for StepFormatter (As Annotation)
- [ ] Add `$argumentName` parsing for CaseDescription
- [ ] Add single `$` count for StepFormatter
- [ ] Add single `$` count for CaseDescription
- [ ] Add tests for StepFormatter
- [ ] Add tests for CaseDescription

cirquit:
Still TODO:

- [ ] write documentation

janschaefer:
What about one of the following names: 
* CaseSummary
* CaseName
* CaseTitle

cirquit:
CaseSummary seems nice because of the __describing__ nature. CaseName and CaseTitle seem to be too specific.

janschaefer:
Also an option woud be CaseAs, because it would be consistent to the As annotation

PR #319, opened 2017-05-27
fix #318

### LINKED ISSUE #318: NullPointerException raised by POJOAnnotationFormatter.format(...)
Hi again, 

Lately, i face a situation where
`POJOAnnotationFormatter.format(java.lang.Object argumentToFormat, Formatf annotation)`
raise a `NullPointerException`.

This occurs when received parameter `argumentToFormat ` is null.
My `POJOAnnotationFormatter` implementation clearly misses something here.
This could be fixed by early inserting a check to test if `argumentToFormat` is null, and if that is the case, immediatly return null which, IMHO, is better than returning an empty string (can't see any indication about this in interface `AnnotationArgumentFormatter.format(java.lang.Object argumentToFormat, Formatf annotation)` contract though.)
By returning null (rather than an empty string), this make this null POJO object further evictable when part of a `@Table` formatting where attribute `includeNullColumns` is set to `false`.

dgrandemange:
- [x] fix POJOAnnotationFormatter to return literal string "null" when given object is null
- [x] complete datatable examples to test this behavior
- [x] migrate datatable examples model/format/annotation classes from src/test/java to src/main/java

### DIFF
--- a/jgiven-core/src/main/java/com/tngtech/jgiven/format/POJOAnnotationFormatter.java
+++ b/jgiven-core/src/main/java/com/tngtech/jgiven/format/POJOAnnotationFormatter.java
@@ -4,6 +4,7 @@ import java.lang.annotation.Annotation;
 import java.lang.reflect.Field;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 
 import com.google.common.base.Function;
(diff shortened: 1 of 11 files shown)


janschaefer:
Thank you for the pull request! The build currently fails, however, because you use the `Objects.isNull` method from JDK 8. JGiven, however, is currently backwards compatible until Java 6. Could you fix that?

dgrandemange:
Hi,

Got it.
Actually, the Gradle Eclipse integration plugin i use didn't show any compile errors in the jgiven-core module. So this explains that.

Anyway, it should be ok by now.

Issue #329, opened 2017-10-17
IllegalArgumentException is thrown when creating thumbnails


maccluca:
An IllegalArgumentException: Width (0) and height (0) cannot be <= 0 is thrown in the Html5AttachmentGenerator when resizing an image to create the corresponding thumbnail.
If the image width and height are small enough, they are rounded to 0 during rescaling.

janschaefer:
this is a good one :-)

PR #330, opened 2017-10-17
Fix IllegalArgumentException if size of thumbnail is rounded to 0

maccluca:
Fixes issue #329.


janschaefer:
Thanks for the fix!

PR #335, opened 2018-02-02
Added simple description helper

nikowitt:
Hi Jan,

some tiny pull request to have the ability to just print out text in the report without triggering any other code which we sometimes need in very simple cases where we don't have a real given, but want to provide some information without having to create a stage method.

Regards,
Niko


janschaefer:
Yes, makes sense. Thanks for the PR.

PR #337, opened 2018-02-12
Added marker annotations @Given, @When, @Then

nikowitt:
Hi Jan,
some other small suggestion to be included in master: When a test scenario with three different stage classes is too heavyweight, but to be able to distinguish given/when/then methods in a simple scenario in a single stage, I have realized that adding and using these annotations for documentation purposes is very helpful.
What do you think?
Best regards,
Niko


janschaefer:
The problem I have with this is that it introduces annotations without any real semantics, which, however, look like they have semantics. In addition, it might be that in the future, it could be the case that we actually want to have annotations like `@Given` to have some semantics. So in good old Java fashion, I consider these annotations to be reserved for the future :-). Another argument is that it is very easy to define these annotations yourself. So it is not needed to have them in the JGiven core.

PR #343, opened 2018-05-14
Fix typo in README

sheamunion:


sheamunion:
Yup, you're right. Updated to fix the actual typo.

Issue #338, opened 2018-02-21
given() returns null if all @Test methods in previous class are @Disabled


wulfsberg:
Using jgiven-junit5, 0.15.3

Given two test classes implementing ScenarioTest or SimpleScenarioTest:
 * Class A has all `@Test` methods `@Disabled`.
 * Class B contains "normal" `@Test` methods.
 
When class B is run after class A in the same test run (say, using Maven), the `given()` call in class B returns null.

This only occurs if A _has_ `@Test` methods which are _all_ `@Disabled`.

E.g.

```
public class DisabledReproA extends ScenarioTest<ReproStage, ReproStage, ReproStage> {
  @Test
  @Disabled("for reasons")
  void testA() {
    given().something_is_set_up();
  }
}

public class DisabledReproB extends ScenarioTest<ReproStage, ReproStage, ReproStage> {
  @Test
  void testB() {
    // This will crash with a nullpointer exception:
    given().something_is_set_up();
  }
}

public class ReproStage extends Stage<ReproStage> {
  public ReproStage something_is_set_up() {
    return self();
  }
}
```

PR #348, opened 2018-05-20
fix issue with JUnit 5 where a test class has only disabled methods, …

### LINKED ISSUE #388: Provide scenarioTests with 2 stages (instead of 3 or 1)
We have been using jgiven for quite a while now and found that we tend to always use an in-between base class which is not <GIVEN, WHEN,THEN> nor <ALL> but <GIVENWHEN, THEN>.

Motivation: when working with tests on multiple state changes, the when-part of the first test becomes part of the given-part of the next ... while the "assertion" (then) stage remains the same. (we use this in process tests and cqrs aggregate tests, where the state of the system-under-test ist dtermined by a sequence of actions).

So we came up with something like

     ActionAssertScenarioTest<ACTION_STAGE, ASSERTION_STAGE> extends ScenarioTest<ACTION_STAGE, ACTION_STAGE, ASSERTION_STAGE>

and use it as base class of our tests.

I would love to have that in jgiven directly as an alternative to the SimpleScenarioTest<ALL>, or working with class hierarchies for GIVEN and WHEN stages.
I think the change is a small one, just provide the mentioned abstract base class.  
What do you think? would you be open for a PR?

janschaefer:
…fixes #388

### DIFF
--- a/jgiven-junit5/src/main/java/com/tngtech/jgiven/junit5/JGivenExtension.java
+++ b/jgiven-junit5/src/main/java/com/tngtech/jgiven/junit5/JGivenExtension.java
@@ -7,19 +7,13 @@ import java.util.ArrayList;
 import java.util.EnumSet;
 import java.util.List;
 
-import com.tngtech.jgiven.config.ConfigurationUtil;
-import com.tngtech.jgiven.impl.Config;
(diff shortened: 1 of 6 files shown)


janschaefer:
fixes #338

PR #353, opened 2018-06-25
various dependencies updates

based2:


janschaefer:
Thanks for the effort, but JGiven seems to not compile with these updates. See Travis output.

codecholeric:
I think you can close this PR, I did the same on ArchUnit :wink:

PR #354, opened 2018-07-05
Allow for partial application of paramaters when using placeholders.

romanticjo:
It seemed weird that JGiven would output every parameter in a formatted @As annotation when only a specific parameter is needed.  This patch addresses that.

All but one test  of the StepFormatter is passing.  This one fails:
https://github.com/TNG/JGiven/blob/master/jgiven-core/src/test/java/com/tngtech/jgiven/report/model/StepFormatterTest.java#L88

```
{ "$$ foo", asList( "bool" ), asList( true ), null, "", "\\$ foo true" },
```
I'm not sure why this is desirable, maybe this is just a side effect of how the replacement logic was implemented in the first place?  If this is the case, can we change the behavior of this test to be:

```
{ "$$ foo", asList( "bool" ), asList( true ), null, "", "\\$ foo" },
```
?

### DIFF
--- a/jgiven-core/src/main/java/com/tngtech/jgiven/report/model/StepFormatter.java
+++ b/jgiven-core/src/main/java/com/tngtech/jgiven/report/model/StepFormatter.java
@@ -153,6 +153,8 @@ public class StepFormatter {
         boolean dropNextWhitespace = false;
         StringBuilder currentWords = new StringBuilder();
 
+        boolean performedReplacement = false;
+
         for( int i = 0; i < stepDescription.length(); i++ ) {
 
             boolean dollarMatch = stepDescription.charAt( i ) == '$';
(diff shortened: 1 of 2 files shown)


janschaefer:
I agree that this behavior actually makes more sense. The reason for this implementation is that initially parameters were always just added to the end. Placeholders were implemented later. And the `@As` annotation came even later. As I did not want to have an inconsistency in the replacement logic, the replacement logic of the `@As` annotation is the same as the replacement logic of placeholders in method names. In addition, this implementation ensures that you do not accidentally forget a parameter in the output, e.g. after a refactoring. This is why parameters that should not be shown in the output have to be annotated with `@Hidden`.

In any case, changing the replacement logic now, would break existing code that expects that behavior.

romanticjo:
Very well, the workaround seems reasonable.  Perhaps it's worth thinking about fixing this when a major version release occurs?  Consider that it breaks the principle of least surprise for new users. Thanks for an overall good library, I'm having fun with it paired with Jspec.

PR #358, opened 2018-07-14
Add support for Spock Framework.

mustaine:
Using ```ScenarioSpec``` allow to use JGiven with Spock Framework.

## Simple example
```
    def "something should happen"() {
        expect:

        given().some_state()
        when().some_action()
        then().some_outcome()
    }
```

## Supports Spock Parameters (looks awesome in the reports)
```
    @Unroll
    def "be able to use tables #param"() {
        expect:
        given().some_state_$(param)
        when().some_action()
        then().some_outcome()

        where:
        param        | _
        "param"      | _
        "word param" | _
    }
```

Thanks @jordi9 for the Spock Spec and the inspiration.


janschaefer:
How cool is that?? I always considered Spock as an JGiven alternative and would never have thought about combining both. Truly amazing!

PR #371, opened 2018-10-06
@JGivenStage enforces prototype mode to ensure a fresh instance 

### LINKED ISSUE #362: Prototype scope with @JGivenStage should be documented or enforced
Hi,

The fact a stage in Spring context (annotated with `@JGivenStage` or other) must have a prototype scope should be documented at least in `JGivenStage` javadoc and JGiven user guide as a reminder.

I don't see any case where a `JGivenStage` shouldn't have prototype scope because `@ProvidedScenarioState` fields are not reinitialized between tests execution and thus given context is messed up by previous test cases.

If there is no reason to have singleton stage, maybe the prototype scope should be enforce when stages are used with Spring.
I don't know if adding annotation `@Scope(ConfigurableBeanFactory.SCOPE_PROTOTYPE)` on `JGivenStage` is enough to activate this behavior or if it must be done in `JGivenBeanFactoryPostProcessor`.

ahus1:
There is added JavaDoc and two test cases. I needed to store a reference to the previous step, so I added a private static field to make it work.

Any ideas how to improve this? Should the documentation be enhanced?

If the answers are "no" it's ready to be merged from my point of view.

closes #362

### DIFF
--- a/jgiven-spring/src/main/java/com/tngtech/jgiven/integration/spring/JGivenStage.java
+++ b/jgiven-spring/src/main/java/com/tngtech/jgiven/integration/spring/JGivenStage.java
@@ -6,10 +6,13 @@ import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 
+import org.springframework.beans.factory.config.ConfigurableBeanFactory;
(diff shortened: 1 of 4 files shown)


janschaefer:
Thanks!

Issue #377, opened 2019-01-06
Update to Gradle 5


mustaine:
Update to Gradle 5 will allow improving dependency management among other things.

PR #378, opened 2019-01-06
Update to Gradle 5.

mustaine:
Issue #377 

Gradle 5: https://gradle.org/whats-new/gradle-5/


mustaine:
Upgrade to Gradle 5.0 will requires Java 8 or higher. 
Should we stop supporting Java 7?

PR #379, opened 2019-01-06
Remove support for Java 7. Only Java 8 or higher.

mustaine:
To support Gradle 5 needs Java 8 or higher.


janschaefer:
I thought long about that and I think I agree. It will make development easier for JGiven to not support Java 7 anymore. In particular, as other frameworks, e.g. Spring 5 also require Java 8 now.

PR #381, opened 2019-01-19
Mustaine update gradle 5

janschaefer:


build.gradle
```diff
@@ -15,6 +15,7 @@ buildscript {
 plugins {
     id 'com.github.ben-manes.versions' version '0.17.0'
     id "org.sonarqube" version "2.5" apply false
+    id 'com.gradle.build-scan' version '1.16'
```
wolfs: I think you should be using build scan plugin `2.x` with Gradle 5.x. The latest build scan plugin version is `2.1`, so I suggest you go with that.

build.gradle
```diff
@@ -25,8 +26,9 @@ apply plugin: 'io.codearte.nexus-staging'
 description = 'JGiven - BDD in plain Java'
 def jacocoEnabled = !JavaVersion.current().isJava11Compatible();
 
-task wrapper(type: Wrapper) {
-    gradleVersion = '4.10.2'
+wrapper {
+    gradleVersion = '5.1'
```
wolfs: Gradle `5.1.1` is the latest version and contains a few bug fixes. I suggest you use this version.

Issue #266, opened 2016-12-23
Add possibility to set up stages


codecholeric:
It would be great if it was possible, to mark a method in a way, that JGiven completely ignores any special handling and doesn't set the stage as active (and hence causes `@AfterStage`-Methods to be called. An use case would be sth. like

```
@Before
public void setUp() {
    secondStage.prepareSomethingThatCantBeProvidedViaStageInjection(..)
}

@Test
public void someTest() {
    firstStage.do_something(); // now @AfterStage of secondStage is called :-(
}
```

PR #385, opened 2019-02-02
Make @DoNotIntercept not trigger stage changes.

codecholeric:
Better late than never :wink: 
This PR changes JGiven's behavior with respect to methods that should not be intercepted (annotated with `@DoNotIntercept` or methods of `java.lang.Object`). These methods will now not trigger a change of the active stage and hence not trigger any `@AfterStage` methods.
I have thought about adding a new annotation to solve my issue #266, but in fact I don't see the use of those methods triggering any JGiven behavior.
This is pretty much the behavior I had expected from `@DoNotIntercept` anyway, to have absolutely no side effects within JGiven.
If I have missed a point here (I don't know the code base very well), let me know. But since all tests seemed to pass, I figured that there was no specific reason for the old behavior.

Resolves: #266

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 ## Backwards Incompatible Changes
 
 * Java 7 is not supported anymore
+* Calling stage methods annotated with `@DoNotIntercept` or declared within `java.lang.Object` will not trigger
+a stage change anymore [#385](https://github.com/TNG/JGiven/pull/385)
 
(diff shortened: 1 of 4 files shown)


codecholeric:
Done :smiley: Finally the label 1.0 :wink: (after having been super stable for such a long time, never had problems in our project!!)

janschaefer:
Thanks!

PR #389, opened 2019-04-09
Apply plugin java-library for dependency management.

mustaine:
As part of updating gradle, required to fix version for `gradle-animalsniffer-plugin`

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -27,7 +27,7 @@ description = 'JGiven - BDD in plain Java'
 def jacocoEnabled = !JavaVersion.current().isJava11Compatible();
 
 wrapper {
-    gradleVersion = '5.3.1'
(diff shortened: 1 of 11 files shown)


mustaine:
@janschaefer looks like the annotation `@FailingOnPurpose` is not working properly. Could you check it?

janschaefer:
There seems to be a problem with the JGiven Gradle plugin as it uses an internal API.

mustaine:
No worries @janschaefer as I need to update the gradle plugin.

mustaine:
I had some issues to make it pass. The problem was that in the main `build.gradle` the task `jgivenHtml5Report` only use `testCompile`.

janschaefer:
Cool. Thanks!

Issue #390, opened 2019-04-09
Upgrade guava version for compatibility with newer plugins


jromero:
https://github.com/TNG/JGiven/blob/9cea246233f63dead47740da1ef495082e6931db/jgiven-core/src/main/java/com/tngtech/jgiven/report/json/JsonModelTraverser.java#L19

This seems to fail along with newer dependencies that use newer versions of [guava](https://google.github.io/guava/releases/24.0-jre/api/docs/com/google/common/io/Files.html#fileTreeTraverser--). Removed in [25.0-jre](https://google.github.io/guava/releases/25.0-jre/api/docs/com/google/common/io/Files.html#fileTraverser--).

**Workaround**
```groovy
buildscript {
  configurations.classpath {
    resolutionStrategy {
      force 'com.google.guava:guava:24.0-jre'
    }
  }
}

plugins {
  id "com.tngtech.jgiven.gradle-plugin" version "0.17.1"
}

```

**Error**
```
org.gradle.api.tasks.TaskExecutionException: Execution failed for task ':spi:jgivenIntegrationTestReport'.
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.executeActions(ExecuteActionsTaskExecuter.java:110)
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.execute(ExecuteActionsTaskExecuter.java:77)
        at org.gradle.api.internal.tasks.execution.OutputDirectoryCreatingTaskExecuter.execute(OutputDirectoryCreatingTaskExecuter.java:51)
        at org.gradle.api.internal.tasks.execution.SkipUpToDateTaskExecuter.execute(SkipUpToDateTaskExecuter.java:59)
        at org.gradle.api.internal.tasks.execution.ResolveTaskOutputCachingStateExecuter.execute(ResolveTaskOutputCachingStateExecuter.java:54)
        at org.gradle.api.internal.tasks.execution.ValidatingTaskExecuter.execute(ValidatingTaskExecuter.java:59)
        at org.gradle.api.internal.tasks.execution.SkipEmptySourceFilesTaskExecuter.execute(SkipEmptySourceFilesTaskExecuter.java:101)
        at org.gradle.api.internal.tasks.execution.FinalizeInputFilePropertiesTaskExecuter.execute(FinalizeInputFilePropertiesTaskExecuter.java:44)
[...]

PR #393, opened 2019-04-21
Update guava version.

mustaine:
Fix issue https://github.com/TNG/JGiven/issues/390


janschaefer:
I had to upgrade some other dependency to be able to build with the new Guava version. Did that in #398.

PR #394, opened 2019-04-22
Add OpenJDK 12 build on Travis.

mustaine:


janschaefer:
Thanks for the PR. It seems that JGiven does not support JDK12 yet. I will work on fixing this.

PR #406, opened 2019-05-10
introduce DualScenarioTest

jangalinski:
see #388 

Motivation: keep steps for given and when in same stage so you can easily build tests that build on each other like 

```
given().doA();
when().doB();
then().assertState();
```

and

```
given().doA().and().doB();
when().doC();
then().assertState();
```

Naming things was the hard part, I came up with DualScenarioTest ... if you are generally fine with the idea, I will continue adding tests and docs.


janschaefer:
Yeah, go ahead :-)

jangalinski:
Do you have any hints for testing? I do not see any explicit tests for the SimpleScenarioTest(Base) ... but basically, I am doing the same with the DualScenarioTest(Base) ... is it ok to skip this?

jangalinski:
I updated the adoc and mentioned `DualScenarioTest` where ever (2 locations) `SimpleScenarioTest` is mentioned already.

jangalinski:
Could you check please if this is enough or give concrete tasks that should be done before this could get in a release?

PR #407, opened 2019-05-10
introduce jgiven-kotlin

jangalinski:
Kotlin is working just fine with jgiven, but in day to day usage, some problems occur. This module provides useful fixes and extensions:

* it contains a JGivenStage annotation that can be used to configure the kotlin "all-open" plugin, otherwise stage classes and all stage methods would have to be explicitly marked as open 
* "when" is a keyword in kotlin, so Stage.when() has to be written as Stage.`when`() ... I introduced extension properties for given()/when()/then()/... so a kotlin test can be written like:

```
class JGivenKotlinExtensionTest : ScenarioTest<JGivenKotlinGiven, JGivenKotlinWhen, JGivenKotlinThen>() {

  @Test
  fun `add two numbers`() {
    GIVEN
        .number(5)

    WHEN
        .we_add_number(7)

    THEN
        .the_result_is(12)
  }
}

```

What do you think?

Attention: this is just a kick-off PR, if you would integrate this into jgiven, we'd have to set up the dokka plugin for java doc generation.

### DIFF
--- a/jgiven-kotlin/build.gradle.kts
+++ b/jgiven-kotlin/build.gradle.kts
@@ -1,0 +1,28 @@
+import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
+
+plugins {
+  kotlin("jvm") version "1.3.21"
+  id("org.jetbrains.kotlin.plugin.allopen") version "1.3.21"
+}
+
+dependencies {
+  api(project(":jgiven-core"))
+
+  implementation(kotlin("stdlib-jdk8"))
+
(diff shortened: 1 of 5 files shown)


jangalinski:
@janschaefer did you have a chance to have a look into this?

janschaefer:
As far as I can evaluate it (I am not a Kotlin developer) it looks good.

jangalinski:
Hi, what is the state of this PR? Any chance to consider support for kotlin for 1.0.0?

PR #426, opened 2019-09-12
Asciidoc multilines

slezier:
Allow a multilines values with empty line to be render as asciidoc literal block.

When one or more empty line is present in values, the word is rendered as literal only for the text before the first empty line. By using the literal block notation (4 dots) all the enclosed value is rendered as literal.


janschaefer:
Thanks!

Issue #418, opened 2019-08-30
dry run option just to generate the report?


adrian-herscu:
Is there any option for running a dry run just to see how the test report would be generated?
Any suggestion about how to implement such thing?

janschaefer:
The `@Pending` annotation could be used for single tests (https://static.javadoc.io/com.tngtech.jgiven/jgiven-core/0.18.0/com/tngtech/jgiven/annotation/Pending.html), but I am not sure if this is what you want.

adrian-herscu:
> The `@Pending` annotation could be used for single tests (https://static.javadoc.io/com.tngtech.jgiven/jgiven-core/0.18.0/com/tngtech/jgiven/annotation/Pending.html), but I am not sure if this is what you want.

Yes, I am already using it for problematic tests.
I am looking for a central hooking point that I can use in order to make all stages temporarily appear as pending, just for one run. Thought about using AspectJ to add the pending annotation at runtime but still not found whether such thing as adding annotations at runtime is supported.

janschaefer:
Ok, we could think about a global configuration setting for that. May I ask what your concrete use case is?

adrian-herscu:
It's a bit complex... I am working for a medical device company. We want automated tests to run nightly for regression purpose. Jenkins+Maven+TestNG+JGiven are fine for that :) We publish the JGiven report and everybody is happy! On the other hand, we cannot prove that the automated tests are valid, i.e. that they do whatever they are supposed to do, and do that without substantial effort. Since these tests are custom made for our product, they change as new features are added/modified, so we would need to validate them for every release. In simple words, we need to show how we test our tests and also do those tests! This effort is definitely greater than just running these tests manually. Still, we need to issue a test document to be executed and signed by manual testers. We already implemented a reporter to do just that. It runs via Maven, gathers all JGiven's JSON files and spits out an HTML document that latter can be imported into Word for additional mingling. But in order to have those JSON files, we still need to actually run these tests. Besides consuming lots of time, it also fails sometimes, requiring a new run. All this just to issue a test document for manual execution!
If there would be a dry run switch, we could get all JSON report files completely and could run our report generator on them :) Hope it makes sense...

janschaefer:
Ok, I think I understand. I also think that it is not difficult to implement, given that this mainly means to execute all scenarios as if they are pending.

janschaefer:
[...]

PR #431, opened 2019-10-23
Dry-run support

mustaine:
Support for dry-run mode via a System property `jgiven.report.dry-run`=`true`
Feature to support #418 

@janschaefer could you give some feedback? 
We would like to use it to generate the narrative from the tests without executing the steps, it might be useful to generate it when the task is avoided because have been cached.


mustaine:
Could we merge https://github.com/TNG/JGiven/pull/428 so the Dependency Managements of Gradle gets improved?

janschaefer:
I fixed the comment above my self in #435

PR #434, opened 2019-11-07
Use gradle plugins dsl

mustaine:
Refactor gradle files

Fix JGiven gradle plugin #437 using gradle 6

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,33 +1,18 @@
-buildscript {
-    repositories {
-        mavenCentral()
-        jcenter()
-    }
-
-    dependencies {
-        classpath 'org.kt3k.gradle.plugin:coveralls-gradle-plugin:2.8.1'
-        classpath 'org.asciidoctor:asciidoctor-gradle-plugin:1.5.3'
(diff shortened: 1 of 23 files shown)


mustaine:
I might be missing something, but I did not see any advantage of moving Gradle towards kotlin at the moment. Also, AFAIK it is slower.

Do you have something in mind?

mustaine:
@janschaefer need the latest commit to upgrade to gradle 6. Otherwise projects using gradle 6 can not use the jgiven gradle plugin

jangalinski:
I personally like the kotlin dsl and use it for all new projects ... I saw a presentation by the gradle inventor once where he stated that had kotlin be around back then,  he would have chosen it from the beginning on ... but except personal preference and the benefit of type safety there is no real killer argument to migrate. 
I was not aware of the performance issue ... from what I learned from 5 seconds google browsing, currently it really is slower due to script compilation (up to 3x on first execution, ~10% on repeated execution) ... good to know ... 

So forget my comment, go ahead ... I am looking forward to using the new gradle plugin asap.

Issue #310, opened 2017-03-20
Use ByteBuddys own cache implementation


janschaefer:
Since v1.6.0, ByteBuddy provides its own cache implementation. This should be used instead of JGiven's own implementation.

janschaefer:
@mustaine sure, there are two classes that implements caching. I think I meant the second one with this ticket, but I cannot 100% remember anymore :-) 

- jgiven-core/src/main/java/com/tngtech/jgiven/impl/util/FieldCache.java
- jgiven-core/src/main/java/com/tngtech/jgiven/impl/CachingStageClassCreator.java

l-1squared:
Apparently this is already resolved

PR #438, opened 2019-11-23
Use ByteBuddy TypeCache instead of custom cache.

mustaine:
#310 Uses `TypeCache` from `ByteBuddy` instead of custom cache.


mustaine:
`FieldCache.java` will neede to use `TypeCache` as well

PR #461, opened 2020-06-02
appended changes since 8.01.2020 to changelog

l-1squared:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,13 +5,13 @@ due to the removal of all deprecated classes and methods and the drop of Java 7 
 
 ## New Features
 
-* Java 12 is supported now [#397](https://github.com/TNG/JGiven/pull/397)
+* Java 13 is supported now [#447](https://github.com/TNG/JGiven/pull/447) (thanks to jsalinaspolo)
 * The `CurrentStep` interface has a new method `setName` to change the name of a step programmatically [#386](https://github.com/TNG/JGiven/issues/386)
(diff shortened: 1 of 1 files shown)


l-1squared:
> Why should Java 9 not be supported anymore? This would be quite bad.

I basically went by this commit and its message: https://github.com/TNG/JGiven/commit/b8409a912563c497129e2a4c21303a80295f9631

jsalinaspolo:
Java 9 is supported, just we are not building in Java9 as we build in Java8, 11 and 13

jsalinaspolo:
Should we use links to the contributors instead of plain text? It would fix if a contributor changes the username - @jsalinaspolo instead of jsalinaspolo ?

l-1squared:
> Should we use links to the contributors instead of plain text? It would fix if a contributor changes the username - @jsalinaspolo instead of jsalinaspolo 

Ah, a misunderstanding on my side. I will change the text accordingly :) And try out the user tags.

l-1squared:
> Should we use links to the contributors instead of plain text? It would fix if a contributor changes the username - @jsalinaspolo instead of jsalinaspolo ?

according to https://help.github.com/en/github/setting-up-and-managing-your-github-user-account/changing-your-github-username, it seems that username links are not protected against a username change. the @jsalinaspolo  you can use here also seem to effectively be converted to a link to your user profile. As such a see no possibility for us to protect the changelog document against username changes.

l-1squared:
@janschaefer is there still something missing?

PR #472, opened 2020-06-17
Implementation of "Filler Words"

richard-stowe:
Hi @janschaefer and @l-1squared.

I have the first PR ready. In addition to the new `@FillerWord` concept I've also added `@Punctuation`, let me know what you think?

Richard

### DIFF
--- a/docs/stages.adoc
+++ b/docs/stages.adoc
@@ -110,6 +110,67 @@ public SELF comma() {
 }
 ----
 
+=== Filler Words
+
+Filler words can be used to build a common and reusable vocabulary for your tests. This enables you to write
+beautifully fluent tests whilst at the same time keeping your stage methods clear and concise.
+
+The `@FillerWord` annotation can also be added to stage methods where you want to maintain the continuation of
+of sentences in reports.
+
+For example, given the following filler words:
+
+[source,java]
+----
+include::{sourcedir}/userguide/ExtendedVocabularyTest.java[tags=fillerWords,indent=0]
+----
+
+You can write test stages such as:
+
+[source,java]
+----
+include::{sourcedir}/userguide/ExtendedVocabularyTest.java[tags=givenFiller,indent=0]
+----
+
+This would generate the following report:
+
+----
+Given the ingredients
+      an egg
+      some milk
+      and the ingredient flour
+----
+
+Filler words can also be joined to other words in sentences. Using `joinToPreviousWord` and `joinToNextWord`
+you can create new words without trailing and leading whitespace respectively. A typical use case for using these
+attributes is adding punctuation to your tests.
+
+For example, given the following filler words:
+
+[source,java]
+----
+include::{sourcedir}/userguide/ExtendedVocabularyTest.java[tags=joiningWords,indent=0]
+----
+
+You could write test stages such as:
+
+[source,java]
+----
+include::{sourcedir}/userguide/ExtendedVocabularyTest.java[tags=givenJoiningWords,indent=0]
+----
+
+This would generate the following report:
+
+----
+Given a (clean) worksurface, a bowl and the ingredients:
+     an egg
+     some milk
+     the ingredient flour
+----
+
+Please note that the annotations `@IntroWord` and `@FillerWord` are mutually exclusive. `@IntroWord` takes precedence
+over `@FillerWord`; where there are occurrences of methods having both annotations `@FillerWord` will be ignored.
+
 === State Sharing
 
 Very often it is necessary to share state between steps.
--- a/jgiven-core/src/main/java/com/tngtech/jgiven/annotation/FillerWord.java
+++ b/jgiven-core/src/main/java/com/tngtech/jgiven/annotation/FillerWord.java
@@ -1,0 +1,32 @@
+package com.tngtech.jgiven.annotation;
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Marks a method which will be used as a preamble to the next non-filler stage method.
+ * In reports, filler words inherit the alignment of the stage method that follows.
+ * Examples of filler words that may be used are the, some, for_a.
+ * domain specific filler words can be created for specific use cases.
+ */
+@Documented
+@Retention( RetentionPolicy.RUNTIME )
+@Target( ElementType.METHOD )
+public @interface FillerWord {
+
+    /**
+     * In reports, filler words marked with joinToPreviousWord will be suffixed to the previous word
+     * in the sentence without whitespace.
+     */
+    boolean joinToPreviousWord() default false;
+
+    /**
+     * In reports, filler words marked with joinToNextWord will be prefixed to the next word
+     * in the sentence without whitespace.
+     */
+    boolean joinToNextWord() default false;
+
+}
--- a/jgiven-core/src/main/java/com/tngtech/jgiven/impl/ScenarioModelBuilder.java
+++ b/jgiven-core/src/main/java/com/tngtech/jgiven/impl/ScenarioModelBuilder.java
@@ -10,6 +10,7 @@ import com.tngtech.jgiven.annotation.CaseAs;
 import com.tngtech.jgiven.annotation.CaseAsProvider;
 import com.tngtech.jgiven.annotation.Description;
 import com.tngtech.jgiven.annotation.ExtendedDescription;
+import com.tngtech.jgiven.annotation.FillerWord;
 import com.tngtech.jgiven.annotation.Hidden;
 import com.tngtech.jgiven.annotation.IntroWord;
 import com.tngtech.jgiven.annotation.IsTag;
@@ -46,7 +47,6 @@ import org.slf4j.LoggerFactory;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.Method;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.Optional;
@@ -71,7 +71,7 @@ public class ScenarioModelBuilder implements ScenarioListener {
      */
     private List<StepModel> nestedSteps;
 
-    private Word introWord;
+    private final SentenceBuilder sentenceBuilder = new SentenceBuilder();
 
     private long scenarioStartedNanos;
 
@@ -131,12 +131,13 @@ public class ScenarioModelBuilder implements ScenarioListener {
         ParameterFormattingUtil parameterFormattingUtil = new ParameterFormattingUtil( configuration );
         List<ObjectFormatter<?>> formatters = parameterFormattingUtil.getFormatter( paramMethod.getParameterTypes(), getNames( arguments ),
             paramMethod.getParameterAnnotations() );
-        stepModel.setWords( new StepFormatter( stepModel.getName(), nonHiddenArguments, formatters ).buildFormattedWords() );
 
-        if( introWord != null ) {
-            stepModel.addIntroWord( introWord );
-            introWord = null;
-        }
+        new StepFormatter( stepModel.getName(), nonHiddenArguments, formatters ).buildFormattedWords()
+            .forEach( sentenceBuilder::addWord );
+
+        stepModel.setWords( sentenceBuilder.getWords() );
+
+        sentenceBuilder.clear();
 
         stepModel.setStatus( mode.toStepStatus() );
         return stepModel;
@@ -154,9 +155,15 @@ public class ScenarioModelBuilder implements ScenarioListener {
 
     @Override
     public void introWordAdded( String value ) {
-        introWord = new Word();
-        introWord.setIntroWord( true );
-        introWord.setValue( value );
+        sentenceBuilder.addIntroWord( value );
+    }
+
+    private void addToSentence( String value, boolean joinToPreviousWord, boolean joinToNextWord ) {
+        if ( !sentenceBuilder.hasWords() && currentStep != null && joinToPreviousWord ) {
+            currentStep.getLastWord().addSuffix( value );
+        } else {
+            sentenceBuilder.addWord( value, joinToPreviousWord, joinToNextWord );
+        }
     }
 
     private void addStepComment( List<NamedArgument> arguments  ) {
@@ -193,6 +200,9 @@ public class ScenarioModelBuilder implements ScenarioListener {
     public void stepMethodInvoked( Method method, List<NamedArgument> arguments, InvocationMode mode, boolean hasNestedSteps ) {
         if( method.isAnnotationPresent( IntroWord.class ) ) {
             introWordAdded( getDescription( method ) );
+        } else if ( method.isAnnotationPresent( FillerWord.class ) ) {
+            FillerWord fillerWord = method.getAnnotation( FillerWord.class );
+            addToSentence( getDescription( method ), fillerWord.joinToPreviousWord(), fillerWord.joinToNextWord() );
         } else if( method.isAnnotationPresent( StepComment.class ) ) {
             addStepComment( arguments );
         } else {
--- a/jgiven-core/src/main/java/com/tngtech/jgiven/impl/SentenceBuilder.java
+++ b/jgiven-core/src/main/java/com/tngtech/jgiven/impl/SentenceBuilder.java
@@ -1,0 +1,54 @@
+package com.tngtech.jgiven.impl;
+
+import com.google.common.collect.ImmutableList;
+import com.tngtech.jgiven.report.model.Word;
+
+import java.util.ArrayDeque;
+import java.util.Deque;
+import java.util.List;
+
+class SentenceBuilder {
+
+    private final Deque<Word> words = new ArrayDeque<>();
+
+    private boolean joinNextWord;
+
+    boolean hasWords() {
+        return !words.isEmpty();
+    }
+
+    void clear() {
+        words.clear();
+    }
+
+    List<Word> getWords() {
+        return ImmutableList.copyOf( words );
+    }
+
+    void addIntroWord( String value ) {
+        words.removeIf( Word::isIntroWord );
+        words.addFirst( new Word( value, true ) );
+    }
+
+    void addWord( Word word ) {
+        addWord( word, false, false );
+    }
+
+    void addWord( String value, boolean joinToPreviousWord, boolean joinToNextWord ) {
+        addWord( new Word( value ), joinToPreviousWord, joinToNextWord );
+    }
+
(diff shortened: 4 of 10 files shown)


jgiven-core/src/main/java/com/tngtech/jgiven/impl/ScenarioModelBuilder.java
```diff
@@ -159,6 +171,18 @@ public void introWordAdded( String value ) {
         introWord.setValue( value );
     }
 
+    private void fillerWordAdded( String value ) {
+        fillerWords.add( new Word( value ) );
+    }
+
+    private void punctuationAdded( String value ) {
+        if ( fillerWords.isEmpty() ) {
```
l-1squared, line 179: I am not entirely certain about the point of this clause here. If I understood it correctly, then `fillerWord` is prepended before the step method words. I.e. if fillerwords were empty but I had punctuation then I would have a dot at the beginning of the sentence. Moreover, the punctuation would be treated as its own word, contrary to how it is handled when there are words present.

jgiven-core/src/main/java/com/tngtech/jgiven/impl/util/PrintWriterUtil.java
```diff
@@ -1,15 +1,24 @@
...
+import java.io.UnsupportedEncodingException;
+
 public class PrintWriterUtil {
     public static PrintWriter getPrintWriter( File file ) {
         try {
-            return new PrintWriter( file, Charsets.UTF_8.name() );
+            return new PrintWriter( file, Charsets.UTF_8.name() ) {
+                @Override
+                public void println() {
+                    write( "\n" );
```
l-1squared, line 19: why is that better than the code of the standard `PrintWriter` class? Also, if there is a reason to circumvent the double checked locking in there, wouldn't you at least want to use `this.lineSeparator` instead of a hardcoded `\n`

jgiven-core/src/test/java/com/tngtech/jgiven/impl/ScenarioModelBuilderTest.java
```diff
@@ -398,6 +398,44 @@ public void the_Description_annotation_on_intro_words_is_evaluated() throws Thro
...
+        StepModel step = getScenario().getScenarioCaseModel().getFirstStep();
+        assertThat( step.getCompleteSentence() )
+            .isEqualTo( "Given there is something filled, another something filled and with something else" );
+    }
+
+    @Test
+    public void punctuation_can_be_used_at_the_start_of_a_sentence() throws Throwable {
+        startScenario( "Scenario with punctuation prefixing a sentence" );
+        given().something()
+        .hyphen().something_else();
```
l-1squared, line 433: A so that's why you wanted to be able to add punctuation at the start of a sentence. 
However, even if we want to allow that. I'd still argue that handling of punctuation should be consistent. i.e. it always be part of another word, or never.

jgiven-examples/src/test/java/com/tngtech/jgiven/examples/pancakes/test/ExtendedVocabularyPanCakeTest.java
```diff
@@ -0,0 +1,93 @@
...
+public class ExtendedVocabularyPanCakeTest extends ScenarioTest<GivenIngredients, WhenCook, ThenMeal> {
+
+    @ProvidedScenarioState
+    private final Cook cook = new PanCakeCook();
+
+    @Test
+    public void a_pancake_can_be_fried_out_of_an_egg_milk_and_flour() {
+
+        given() .ingredients().comma() .consisting_of()
+            .asterisk() .an() .egg()
```
l-1squared, line 29: in this case, I'd argue that `asterisk` is functionally closer to a word than to punctuation. But I get now, why you implemented it the way you did. I'm just not completely satisfied...

jgiven-core/src/main/java/com/tngtech/jgiven/impl/util/PrintWriterUtil.java
richard-stowe: Ah, this is something I quickly did to get the build working on my windows machine. I'll revert and look for something a little less "hacky"

jgiven-examples/src/test/java/com/tngtech/jgiven/examples/pancakes/test/ExtendedVocabularyPanCakeTest.java
richard-stowe: Ok, agreed. Maybe I overdid this one. I'll wind back the use of punctuation a little.

jgiven-core/src/main/java/com/tngtech/jgiven/annotation/FillerWord.java
```diff
@@ -0,0 +1,20 @@
...
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Marks a method which will be used as a preamble to the next non-filler stage method.
+ * In reports filler words inherit the alignment of the stage method that follows.
```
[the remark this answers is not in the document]
richard-stowe: Addressed

jgiven-core/src/main/java/com/tngtech/jgiven/annotation/Punctuation.java
```diff
@@ -0,0 +1,32 @@
...
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Marks a method which will be used to represent punctuation in tests.
+ *
+ * <p>
```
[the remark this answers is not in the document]
richard-stowe: Example here has been removed

jgiven-core/src/main/java/com/tngtech/jgiven/impl/ScenarioModelBuilder.java
```diff
@@ -46,13 +48,16 @@
...
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.Optional;
 import java.util.Set;
 import java.util.Stack;
 
+import static com.google.common.collect.Iterables.concat;
+import static com.google.common.collect.Iterables.getLast;
```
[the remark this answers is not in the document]
richard-stowe: Muscle memory pulled in the static imports. I've reverted this and aligned new usages of `google.common.collect` with the previous implementation

jgiven-core/src/main/java/com/tngtech/jgiven/report/model/Word.java
```diff
@@ -78,7 +78,11 @@ public void setValue( String name ) {
     }
 
     public void append( CharSequence word ) {
-        setValue( getValue() + " " + word );
+        append( word, " " );
+    }
+
+    public void append( CharSequence word, CharSequence delimiter ) {
```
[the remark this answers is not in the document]
richard-stowe: I've incorporated your suggestion

jgiven-core/src/main/java/com/tngtech/jgiven/impl/ScenarioModelBuilder.java
```diff
@@ -159,6 +171,18 @@ public void introWordAdded( String value ) {
         introWord.setValue( value );
     }
 
+    private void fillerWordAdded( String value ) {
```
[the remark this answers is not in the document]
richard-stowe: Addressed

jgiven-core/src/test/java/com/tngtech/jgiven/impl/ScenarioModelBuilderTest.java
richard-stowe: If we observe punctuation without the use of filler words what is the desired behaviour? Ignore the punctuation and log a warning?

richard-stowe: Just a thought. How about scrapping `@Punctuation` altogether and having an attribute on `@FillerWord` which allows you to mark it as punctuation, obviously defaulting to false.

If you want to start a sentence with a punctuation filler you may.

l-1squared: I like that idea. Functionally punctuation and filler words are already very similar...

l-1squared: coming to think of it, maybe you want to have that option for every word... then you can do all kinds of crazy sentence building... However, there is no need to get to such a thing right away.

richard-stowe: Yes, my thinking exactly. I'll sort that out.

richard-stowe: > coming to think of it, maybe you want to have that option for every word... then you can do all kinds of crazy sentence building... However, there is no need to get to such a thing right away.

Just image the power of that combined with Navigable Stages. Oh yes!

richard-stowe: @l-1squared I have now incorporated the punctuation concept into filler words. I'm definitely happier with it now. What do you think?

docs/stages.adoc
```diff
@@ -110,6 +110,63 @@ public SELF comma() {
...
+include::{sourcedir}/userguide/ExtendedVocabularyTest.java[tags=givenPunctuation,indent=0]
+----
+
+This would generate the following report:
+
+----
+Given a clean worksurface, a bowl and the ingredients
+     an egg
+     some milk
+     the ingredient flour
```
l-1squared, line 167: (optional) you might a `.` here and in the test just have a better visibility. I actually had to look for the punctuation. Might be an due to the page not being rendered :smile:

jgiven-core/src/main/java/com/tngtech/jgiven/impl/ScenarioModelBuilder.java
l-1squared: --obsolete due to removal of the `@Punctuation` annotation

l-1squared:
keeping this open for another day or say to give @janschaefer another opportunity to voice his opinion

docs/stages.adoc
richard-stowe: I've refined the punctuation concept even further. You may now use punctuation to end an sentence. Hopefully this example conveys the use of punctuation more clearly now.

richard-stowe:
```
   Given a clean worksurface, a bowl and the ingredients:
         an egg
         some milk
         the ingredient flour
```
I think it looks quite smart! :smiley:

janschaefer:
I think this is a great addition to JGiven, so thanks for that! I have two things:

1. Have you verified that this works as expected in the HTML report as well? 

2. I personally had named the attribute punctuation of the FillerWord more technical. Something like: `noWhitespace` or just `whitespace` which is true by default. Because there might be situations in which you maybe have a filler word that is no punctuation and still want to have no white space.

richard-stowe:
@janschaefer, good point. I'm not completely sold on `whitespace` though, how about `affix`? 

image

richard-stowe:
HTML reports look good:

image

richard-stowe:
> @janschaefer, good point. I'm not completely sold on `whitespace` though, how about `affix`?
> 
> image

I've implemented this now. Let me know what you think.

jgiven-core/src/main/java/com/tngtech/jgiven/annotation/FillerWord.java
```diff
@@ -0,0 +1,26 @@
...
+@Documented
+@Retention( RetentionPolicy.RUNTIME )
+@Target( ElementType.METHOD )
+public @interface FillerWord {
+
+    /**
+     * In reports, filler words marked with affix will be appended to the previous filler word
+     * in the sentence without whitespace.
+     */
+    boolean affix() default false;
```
l-1squared, line 24: fine by me. Maybe `affixToWord` for some additional clarity

richard-stowe: After a bit of playing around tests I opted for the use of affix "noun" rather than the "verb". So a filler word can be used as an affix rather than being "affixed" to a sentence. More than happy to switch it if the noun doesn't quite feel right.

l-1squared: Maybe that meaning just doesn't come that naturally to me...If we see it as a noun then we could name the field `isAffix` just to highlight that it is a noun

richard-stowe: The more I think about it the less I like "afffix"! I'll have a think over the weekend how I can better convey the behaviour. We may arrive back at "whitespace".

codecholeric: I'm not completely sold on "affix" either, because personally I would have to search online to get the exact meaning :joy: 
Also if you're very strict, then "an addition to the base form or stem of a word in order to **modify its meaning** or **create a new word**" is not satisfied for the use case you're thinking about, is it? :thinking: Adding punctuation neither modifies the meaning nor creates a new word, so it kind of hides the intention it was originally created for, right?
`whitespace` is technical enough to not have this, but with `whitespace` my problem is that it does not tell you "before or after", does it? :thinking: I might also think then "either it has a whitespace before and after or neither before nor after" which is not true for punctuation either, right?
Don't have a great idea here, but something like `whitespaceBefore` or `joinToLastWord` would at least make it clear that it only refers to the word before this filler word. Or you just support the two use cases of normal words and punctuation and have an attribute like `punctuation`. Which would also match to names on `Word` like `addWord` and `addPunctuation`.

codecholeric:
BTW: Those "addressing feedback", "rename x" of "fixed type" commits I would squash onto the commits where the original change was introduced. Don't think they bring much value in the `master` history, right?

jgiven-core/src/main/java/com/tngtech/jgiven/annotation/FillerWord.java
janschaefer: I agree, I have never heard the word affix before, so maybe not optimal for non-native speakers ;-). Regarding that it is unclear where the space is added, I think this can be clarified in the JavaDoc comment. As space is the default, we could also just called it `noSpace`.

janschaefer: Another alternative would be to have an attribute `prefix` which would be a String instead of a boolean. By default it would be `" "` and can be changed to `""`. But it could also be anything else, like `"-"` for example. I am not sure, however, whether there is any useful use-case for that.

richard-stowe: My current thoughts are along the lines of `whitespaceBefore` and `whitespaceAfter` allowing you to nicely use parentheses if you so desire.

richard-stowe: Or `joinToPreviousWord` and `joinToNextWord`.

richard-stowe: `affix` has been removed. We now have `joinToPreviousWord` and `joinToNextWord` implemented. Supporting both obviously add a bit more complexity but I think the having the ability to join to both previous and next words is a nice addition. Again, all feedback welcomed.

l-1squared: I'll have a look at it first thing tomorrow morning

docs/stages.adoc
```diff
@@ -110,6 +110,64 @@ public SELF comma() {
...
+[source,java]
+----
+include::{sourcedir}/userguide/ExtendedVocabularyTest.java[tags=givenFiller,indent=0]
+----
+
+This would generate the following report:
+
+----
+Given the ingredients
+      an egg
```
l-1squared, line 139: question: If newlines are produced by `@IntroWord` and `an()` is not an intro word, how do we get this sentence on its own line?

jgiven-core/src/test/java/com/tngtech/jgiven/GivenTestStep.java
```diff
@@ -132,12 +147,72 @@ public GivenTestStep table_as_parameter(@Table String[] params) {
...
     }
 
-    @As( "another description" )
     @IntroWord
+    @As( "another description" )
     public GivenTestStep an_intro_word_with_an_as_annotation() {
         return self();
     }
 
+    @FillerWord
```
l-1squared, line 156: what happens if someone has the genius idea of stacking `@FillerWord` and `@IntroWord`? We should make sure that that behavior is well defined and known.

docs/stages.adoc
richard-stowe: `@IntroWord` does not create new lines, it capitalises the word and adds it to the start of the sentence.

In this instance the new lines are created by the `ingredients` and `egg` stage methods.

jgiven-core/src/test/java/com/tngtech/jgiven/GivenTestStep.java
richard-stowe: In this situation '@FillerWord' will be ignored. I've added clarification of this to the docs.

jgiven-core/src/main/java/com/tngtech/jgiven/impl/SentenceBuilder.java
```diff
@@ -0,0 +1,54 @@
+package com.tngtech.jgiven.impl;
+
+import com.google.common.collect.ImmutableList;
+import com.tngtech.jgiven.report.model.Word;
+
+import java.util.ArrayDeque;
+import java.util.Deque;
+import java.util.List;
+
+class SentenceBuilder {
```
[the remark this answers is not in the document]
richard-stowe: I've considered the `SentenceBuilder` to be part of the `ScenarioModelBuilder` "unit". All permutations of `addWord` are tested within the `ScenarioModelBuilderTest`, see coverage report below:

image

I'm more than happy to add more focused tests for the `SentenceBuilder` if you think it'd add value/better clarity of the behaviour.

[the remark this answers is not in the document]
l-1squared: nah, indeed `ScenarioModelBuilder` is probably the better unit to test

PR #480, opened 2020-07-20
Enable automatic OSS-Release

NilsOliverLinden:
The changes needed to enables an automatic OSS-Release via Github include
- Provide credentials from environment variables as there is no gradle.properties
- When triggered by a CI workflow the release should always happen.
- gradle plugin test should use the artifact on the oss.sonatype staging repo similar to the maven plugin test


NilsOliverLinden:
Missing signing. After rework of commit message I will create a new PR

PR #481, opened 2020-07-20
Enable automatic OSS-Release

NilsOliverLinden:
The changes needed to enables an automatic OSS-Release via Github include

- Provide credentials from environment variables as there is no gradle.properties
- When triggered by a CI workflow the release should always happen.
- gradle plugin test should use the artifact on the oss.sonatype staging repo similar to the maven plugin test

### DIFF
--- a/example-projects/junit5/build.gradle
+++ b/example-projects/junit5/build.gradle
@@ -1,10 +1,18 @@
-plugins {
-  id "com.tngtech.jgiven.gradle-plugin" version "${version}"
+buildscript {
+    repositories {
+        if (project.hasProperty('staging')) {
+            maven { url "https://oss.sonatype.org/content/repositories/staging/" }
+        }
+        mavenLocal()
+        mavenCentral()
+    }
+    dependencies {
+        classpath "com.tngtech.jgiven:jgiven-gradle-plugin:${version}"
+    }
 }
 
+apply plugin: 'com.tngtech.jgiven.gradle-plugin'
 apply plugin: 'java'
-apply plugin: 'idea'
-apply plugin: 'eclipse'
 
 compileJava {
     sourceCompatibility = 1.8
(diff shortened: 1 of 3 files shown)


l-1squared:
@NilsOliverLinden  somehow, your commit messages got messed up. Could you double check that?

example-projects/junit5/build.gradle
```diff
@@ -1,10 +1,18 @@
...
+buildscript {
+    repositories {
+        if (project.hasProperty('staging')) {
+            maven { url "https://oss.sonatype.org/content/repositories/staging/" }
+        }
+        mavenLocal()
+        mavenCentral()
+    }
+    dependencies {
+        classpath "com.tngtech.jgiven:jgiven-gradle-plugin:${version}"
```
l-1squared, line 10: @NilsOliverLinden  since the `example-projects` are not linked to the main jgiven project. Does it really make sense to state a classpath dependency here? 

I.e. If I check out the `junit5` project alone, without JGiven, would it still work?

NilsOliverLinden: At least that is the legacy way to implement the plugin. 
See https://plugins.gradle.org/plugin/com.tngtech.jgiven.gradle-plugin

I need the legacy way to be able to specify the url for the staging test.

aaschmid: @l-1squared the dependency is required to be able to use `apply plugin: 'com.tngtech.jgiven.gradle-plugin'`. Otherwise the plugin would not be know to the Gradle build. It replaces the 
```
plugins {
  id "com.tngtech.jgiven.gradle-plugin" version "${version}"
}
``` 
which does all at once but only looks at https://plugins.gradle.org/m2/ for plugins. 

With this change we are able to take the JGiven Gradle plugin from the closed Maven Staging repository and test with the example project before it is actually released to https://plugins.gradle.org/ (and their maven repo).

PR #482, opened 2020-07-21
Issue 478 maven plugin

l-1squared:


build.gradle
```diff
@@ -258,86 +237,80 @@ configure(subprojects) {
...
+        doFirst {
+            if (!isCI && isReleaseVersion && !JavaVersion.current().isJava8Compatible()) {
+                throw new GradleException("Releases have to be done with Java 8")
+            }
+        }
+    }
+    publishing {
+        publications {
+            maven(MavenPublication) {
+                from components.java
```
l-1squared, line 253: This currently fails the `jgiven-android` build :(

l-1squared: not really, because I duplicated everything

scripts/release.sh
```diff
@@ -33,7 +37,7 @@ if [ -n "$(git status --porcelain)" ]; then
 fi
 
 echo Building, Testing, and Uploading Archives...
-./gradlew --no-parallel clean test install uploadArchives
+./gradlew --no-parallel clean test publishToMaven -Pstaging
```
NilsOliverLinden, line 40: Currently, I do not see why you need the staging profile here. In the project there is no staging profile defined, is it?
Only in the gradle-example projects exist a staging profile to determine from where the plugin should be loaded.

build.gradle
```diff
@@ -10,13 +10,11 @@ plugins {
 apply plugin: 'project-report'
 
 description = 'JGiven - BDD in plain Java'
-def jacocoEnabled = !JavaVersion.current().isJava11Compatible();
-
-
+def jacocoEnabled = !JavaVersion.current().isJava11Compatible()
 
 wrapper {
```
aaschmid, line 15: `wrapper` section should not be needed and is not part of at least my repository anymore, see e.g. https://github.com/aaschmid/gradle-cpd-plugin/blob/master/build.gradle

build.gradle
```diff
@@ -43,6 +41,7 @@ allprojects {
 }
 
 
+
```
aaschmid, line 44: Optional: get rid of too many newlines ;-)

build.gradle
```diff
@@ -78,18 +77,81 @@ subprojects {
...
         hsqldbVersion = '2.3.2'
         byteBuddyVersion = '1.10.2'
-        androidGradleVersion= '3.1.0'
-        androidCompileSdkVersion= 27
-        androidBuildToolsVersion= '27.0.2'
+        androidGradleVersion = '3.1.0'
+        androidCompileSdkVersion = 27
+        androidBuildToolsVersion = '27.0.2'
+
+        publishVariant = {String variant->
```
aaschmid, line 84: Maybe add spaces -> `{ String variant ->`

build.gradle
```diff
@@ -221,13 +265,14 @@ configure(subprojects.findAll { !it.name.contains("android") }) {
     }
 
     copyAsciiDoc.finalizedBy(asciidoctor)
+
+    project.ext.publishVariant('java')
```
aaschmid, line 269: I still find it a bit confusing to use `ext` here. 
But haven't a better idea now.

scripts/release.sh
```diff
@@ -18,6 +18,10 @@ if [[ ! $1 =~ ^[0-9]*\.[0-9]*\.[0-9]*(-[A-Z0-9]*)?$ ]]; then
     exit 1
 fi
 
+if [ -z $ANDROID_SDK_ROOT ]; then
+    echo "Android package will NOT be released"
```
aaschmid, line 22: Shouldn't we abort rather abort the release as you want to normally release it together?

settings.gradle
```diff
@@ -52,16 +52,17 @@ include ':jgiven-core',
...
 def release = System.env.RELEASE == "true"
 def android = System.env.ANDROID == "true"
+boolean hasAndroidSourcePath = ( System.env.ANDROID_SDK_ROOT )
 
 if (JavaVersion.current().isJava9Compatible()) {
     include ':jgiven-java9-tests'
 }
 
-if (android || release) {
+if ((android || release) && hasAndroidSourcePath) {
```
aaschmid, line 61: Is this excluded for developers which have no Android SDK?

l-1squared: err yes. After all if you attempt to build everything but don't have that SDK path set, then the android build will throw a beatiful error

scripts/release.sh
l-1squared: yeah, that might be better

build.gradle
l-1squared: giving it some thought, I guess you're right. After all this section only replicates the information that we have already checked in in the `gradle-wrapper.properties` file.

scripts/release.sh
l-1squared: yes. Doing that rebase that late in the day probably wasn't a good idea

PR #484, opened 2020-07-24
Fix and adapt release script

NilsOliverLinden:
- fix bug
- also test maven plugin after release not only gradle plugin
- push created tag to GitHub repository


scripts/release.sh
```diff
@@ -51,28 +51,33 @@ echo Testing Gradle plugin...
...
+  git push $(git config --get remote.origin.url) $VERSION_PREFIXED
 }
+
+echo $CI
+if [[ $CI == "true" ]]
+then
+  releaseRepositoryAndPushVersion
+else
+  read -p "Do you want to release and push now? [y/N]" -n 1 -r
+  if [[ $REPLY =~ ^[Yy]$ ]]
```
l-1squared, line 79: please fix the formatting of these `if` blocks

scripts/release.sh
```diff
@@ -51,28 +51,33 @@ echo Testing Gradle plugin...
...
-fi
-fi
-
 releaseRepositoryAndPushVersion()
 {
   echo Releasing the repository...
   ./gradlew releaseRepository
 
-  echo Publishing Gradle Plugin to Gradle Plugin Repository...
+  echo Testing Maven plugin...
```
l-1squared, line 59: I would opt for a new name here, because that step sound awfully familiar to the "testing Maven plugin" step that comes before. I really had to look, how they are different

PR #498, opened 2020-09-30
Issue 478 handle transient dependencies properly

l-1squared:


l-1squared:
tested the junit5 example against the local maven repository with

fudler (approved):

PR #501, opened 2020-10-09
Improve local release script

l-1squared:


scripts/release_functions.sh
```diff
@@ -0,0 +1,30 @@
...
+          gradle_properties="${gradle_properties} ${command_line_argument}"
+        fi
+        printf "%s" "${gradle_properties}"
+    done
+}
+
+function releaseRepositoryAndPushVersion()
+{
+  echo Releasing the repository...
+  ./gradlew releaseRepository || return 21
```
fudler, line 16: Just out of interest - is there some overview of all those return codes?

l-1squared: no. 
I just have accustomed myself to doing that. If you look at the main script it passes on the error value that was returned by the function. Hence. when you read out the return value can tell  exactly at which line the script failed :)

There might actually be a better solution to achieve the same thing out there, but I never looked

fudler (approved):

PR #502, opened 2020-10-09
Issue-478: also publish artifacts to local repository on release

l-1squared:
Currently the maven plugin build fails, because its maven build tries to collect the jgiven artifacts of the version being released from maven central, which cannot go well. If there is a copy of them in the local repo the build should work


l-1squared:
Jup, just tested it. Publishing to the local repository makes maven find the desired artefacts

PR #505, opened 2020-10-22
Issue-478: ensure the property finders output is handled correctly

l-1squared:
It turns out that there is no way to return a string or an array with elements that themselves contain spaces in a way that bash will interpret the elements correctly (i.e. var1, var 2, var3 -> var1, var, 2, var3). The solution is to use bash's nameref feature. This allows a function to write to a variable directly without knowing the actual variable name.

### DIFF
--- a/scripts/release_functions.sh
+++ b/scripts/release_functions.sh
@@ -1,12 +1,12 @@
 #!/usr/bin/env bash
 
 function find_gradle_property(){
+    [[ "$(declare -p $1 2>/dev/null)" =~ "declare -a" ]] || return 11
+    local -n output_target=$1
     for command_line_argument in "$@"; do
-        gradle_properties=""
         if [[ "${command_line_argument}" =~ ^-P.*=.* ]];then
-          gradle_properties="${gradle_properties} ${command_line_argument}"
+          output_target+=("${command_line_argument}")
         fi
-        printf "%s" "${gradle_properties}"
     done
 }
 
--- a/scripts/test_release_functions.sh
+++ b/scripts/test_release_functions.sh
@@ -4,18 +4,55 @@ SCRIPT_LOCATION=$(dirname -- "$(readlink -f -- "${BASH_SOURCE[0]}")")
 # shellcheck source=./release_functions.sh
 source "${SCRIPT_LOCATION}"/release_functions.sh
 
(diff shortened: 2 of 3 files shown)


l-1squared:
checked that adding gradle properties, especially values with spaces now works correctly

scripts/release_functions.sh
```diff
@@ -1,12 +1,12 @@
 #!/usr/bin/env bash
 
 function find_gradle_property(){
+    [[ "$(declare -p $1 2>/dev/null)" =~ "declare -a" ]] || return 11
+    local -n output_target=$1
     for command_line_argument in "$@"; do
-        gradle_properties=""
         if [[ "${command_line_argument}" =~ ^-P.*=.* ]];then
-          gradle_properties="${gradle_properties} ${command_line_argument}"
+          output_target+=("${command_line_argument}")
```
fudler, line 8: Just to get more insight: what is the point in constructing output_target? Where/how is it used?

scripts/test_release_functions.sh
```diff
@@ -4,18 +4,55 @@ SCRIPT_LOCATION=$(dirname -- "$(readlink -f -- "${BASH_SOURCE[0]}")")
...
+  expected_output="-Pkey=a name"
+  local -a output_array
+  find_gradle_property release.sh 1.0.0 "${expected_output}"
+  printf "Expected \"%s\", got \"%s\"" "${expected_output}" "${output}"
+  [ "${output_array[@]}" == "${expected_output}" ] || return 1 && return 0
+}
+
+test_properties_are_passed_on_correctly(){
+  local -a properties
+  find_gradle_property properties release.sh 1.0.0 -Pkey="a value" -Pindentifier=identifiee
```
fudler, line 47: Should this really be "identifiee"?

scripts/release_functions.sh
l-1squared: In a sense. What we are doing is that we pass the name of the actual variable to the function. Then we create a reference (like an always associated pointer) to that variable with declare -n. And then we address the actual variable by its reference in the function. 

Since bash functions cannot really return complex information, this is the best way to isolate functionality.

scripts/test_release_functions.sh
l-1squared: Ok, I was willing to joke to much again. I will assign boring names.

PR #512, opened 2020-11-10
Changed java9 to junit 5 in releasRepositoryAndPushVersion to avoid f…

fudler:
Changed java9 to junit 5 in releasRepositoryAndPushVersion to avoid failing gradle test

Signed-off-by: Christian Oertel <christian.oertel@tngtech.com>


l-1squared (approved):
Uses the gradle plugin. Looking good

PR #513, opened 2020-11-24
Update versions and timeout

l-1squared:


fudler (approved):
Wow, the versions seemed quite inconsistently outdated.

l-1squared:
just changed the order of commits. Hopefully the validation works now

fudler (approved):

PR #514, opened 2020-11-25
Update versions and timeout

l-1squared:
same as PR #513 but I needed to retrigger the status checks


fudler (approved):
Let's hope this time it does a better job!

PR #605, opened 2021-03-01
Release/v1.0.0

l-1squared:


l-1squared:
actual release was not successful, because of Sonatype issues :(

PR #630, opened 2021-04-19
Gradle 7 demonstration

l-1squared:


jgiven-gradle-plugin/src/main/java/com/tngtech/jgiven/gradle/JGivenPlugin.java
```diff
@@ -36,9 +37,9 @@ private void applyTo( Test test ){
         final JGivenTaskExtension extension = test.getExtensions().create( "jgiven", JGivenTaskExtension.class );
         final Project project = test.getProject();
         ( (IConventionAware) extension ).getConventionMapping().map( "resultsDir",
-                (Callable<File>) () -> project.file( project.getBuildDir() + "/jgiven-results/" + testName ));
+                (Callable<Provider<File>>) () ->  project.provider(() ->project.file( project.getBuildDir() + "/jgiven-results/" + testName )));
```
l-1squared, line 40: I think the problem is somewhere here...

PR #639, opened 2021-04-27
Fix jgiven spock setup

l-1squared:
Effectively reverted the changes from

### DIFF
--- a/jgiven-spock/build.gradle
+++ b/jgiven-spock/build.gradle
@@ -8,10 +8,10 @@ description = "Module for writing JGiven tests with Spock"
 
 dependencies {
     api project(':jgiven-junit')
-    implementation 'org.codehaus.groovy:groovy:3.0.8'
-    implementation 'org.spockframework:spock-core:2.0-M4-groovy-3.0'
-    implementation "org.spockframework:spock-junit4:2.0-M5-groovy-3.0"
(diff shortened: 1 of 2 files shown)


l-1squared:
looking at the log output it seems like groovy2 and java 13 don't get along too well....

jsalinaspolo (dismissed):
LGTM

jgiven-spock/build.gradle
```diff
@@ -8,10 +8,9 @@ description = "Module for writing JGiven tests with Spock"
 
 dependencies {
     api project(':jgiven-junit')
-    implementation 'org.codehaus.groovy:groovy:3.0.8'
-    implementation 'org.spockframework:spock-core:2.0-M4-groovy-3.0'
-    implementation "org.spockframework:spock-junit4:2.0-M5-groovy-3.0"
-    testCompile project(':jgiven-html5-report')
+    implementation 'org.codehaus.groovy:groovy:2.5.13'
+    implementation('org.spockframework:spock-core:1.3-groovy-2.5')
+    testImplementation project(':jgiven-html5-report')
```
fudler, line 13: As far as I know brackets shouldn't make any difference, or? In this case we should stick to a unified notion.

fudler (approved):

PR #685, opened 2021-07-01
Create github actions file to run on PRs and branch pushes

l-1squared:
This replaces travis and will serve to validate future pushes.


.github/workflows/jgiven_branch_and_pr_build.yml
```diff
@@ -0,0 +1,27 @@
+name: Java CI with Gradle
+on: [push, pull_request]
+
+jobs:
+  build:
+    name: Build JGiven on Java ${{ matrix.Java }} with Junit ${{matrix.junit}}
```
fudler, line 6: Is it necessary to use `${{matrix.java}}` instead of `${{ matrix.*J*ava }}` here?

.github/workflows/jgiven_branch_and_pr_build.yml
```diff
@@ -0,0 +1,27 @@
...
+          java-version: ${{matrix.java}}
+          distribution: 'adopt'
+      - name: Grant execute permission for gradlew
+        run: chmod +x gradlew
+      - name: Build with Gradle  and Junit ${{matrix.junit}}
+        env:
+          JUNIT_VERSION: ${{matrix.junit}}
+        run:   "./gradlew -i -s check -DjunitVersion=${JUNIT_VERSION}"
+      - name: Run Jacoco
+        if: ${{matrix.java == '8' && github.ref == 'refs/heads/master' && github.head_ref == null}}
```
fudler, line 26: This might be due to the general syntax of .yml files, but somehow this condition seems odd - shouldn't it be something like 

`if: (${{matrix.java}} == '8' && blabla )` ?

fudler (changes requested):

.github/workflows/jgiven_branch_and_pr_build.yml
l-1squared: github doesn't seem to care about capitalization. But you're right. I should be consistent about it

.github/workflows/jgiven_branch_and_pr_build.yml
l-1squared: https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions recommends using this syntax to have this evaluated as expression

fudler: This is something one needs to get used to, but in this case it's fine.

fudler (approved):

PR #686, opened 2021-07-02
Fix/remove symlinks from example projects

l-1squared:


scripts/test_helper_functions.sh
```diff
@@ -0,0 +1,26 @@
...
+  echo 'version=0.x' > "${test_file}"
+  update_version '1.0' "${test_file}"
+  grep 'version=1.0' "${test_file}"
+  return $?
+}
+
+test_update_version_fails_if_property_not_in_file() {
+  local test_file="properties.tmp"
+  trap "rm -f ${test_file}" SIGTERM SIGINT
+  echo 'venison=0.x' > "${test_file}"
```
fudler, line 19: Typo

fudler (changes requested):

scripts/test_helper_functions.sh
l-1squared: you mean "venison"? That is actually on purpose as I want to see that I fail hard if I try to update a property that doesn't exist

It is meant as a safeguard to ensure that the example projects always get updated, b.c. I cannot release anything if I happen to just change the property name in one of them

fudler (approved):

fudler (approved):

PR #696, opened 2021-07-21
Using a single frontend stage that acts as three stages in the backend

andru47:


jgiven-core/src/main/java/com/tngtech/jgiven/impl/ScenarioExecutor.java
```diff
@@ -50,6 +51,7 @@
     private Object currentTopLevelStage;
     private State state = State.INIT;
     private boolean beforeScenarioMethodsExecuted;
+    private static final ThreadLocal<String> lastExecutedStageClassName = new ThreadLocal<>();
```
l-1squared, line 54: does this need to be a constant?

jgiven-core/src/main/java/com/tngtech/jgiven/impl/ScenarioExecutor.java
```diff
@@ -260,6 +266,22 @@ StageState getStageState(Object stage) {
         return stageState;
     }
 
+    private String getStageNameOrGiven(Object stage, String lastExecutedClassName) throws NoSuchFieldException, IllegalAccessException {
+        String name = SingleStageNameFieldGetterSetter.getStageName(stage);
+
+        return name == null ? "GIVEN" : name;
```
l-1squared, line 272: we need to test this against a scenario where 
```
class SomeTest {

  // copy Scenario methods here

  @ScenarioStage
  private Stage<T> notGiven;

  @Test
  public void test(){
    notGiven.I_run_a_nonconformant_JGiven_test();
  }
}

Comment (l-1squared) [reply jgiven-core/src/main/java/com/tngtech/jgiven/impl/ScenarioExecutor.java:272]: a general problem I have with this setup is that we are using a magic keyword here  to make things work.  A casual reader does not know why, of all things, "GIVEN" is the default here.

File: jgiven-core/src/main/java/com/tngtech/jgiven/impl/Scenario.java [hunk 3]
```diff
@@ -30,14 +32,17 @@ public Scenario( Class<GIVEN> givenClass, Class<WHEN> whenClass, Class<THEN> the
     }
 
     public GIVEN getGivenStage() {
+        SingleStageNameFieldGetterSetter.setStageName(givenStage, "GIVEN");
```
l-1squared, line 35: weird place to perform a setter operation. could this be done earlier?

jgiven-core/src/main/java/com/tngtech/jgiven/impl/util/ReflectionUtil.java
```diff
@@ -1,14 +1,13 @@
 package com.tngtech.jgiven.impl.util;
 
+import static java.lang.String.format;
```
l-1squared, line 3: protip: separate formatting changes from functional changes.  (at least in a separate commit), especially if they cover the whole file.  This eases the reviewing process.

jgiven-core/src/main/java/com/tngtech/jgiven/impl/util/ReflectionUtil.java
```diff
@@ -180,109 +164,134 @@ public static void invokeMethod( Object object, Method method, String errorDescr
...
+    public static void setField(String fieldName, Object object, Object value, String errorDescription) {
         try {
-            object.setAccessible( true );
-        } catch( SecurityException e ) {
-            log.debug( "Caught exception: ", e );
-            log.warn( "Could not make {} accessible, trying to access it nevertheless and hoping for the best.",
-                    toReadableString( object ), errorDescription );
+            setField(object.getClass().getDeclaredField(fieldName), object, value, errorDescription);
+        } catch (NoSuchFieldException e) {
+            log.debug(
```
l-1squared, line 247: since we use SLF4J, you don't need to use `String::format`, you can do `log.debug("a message with inserted values {},{}", value1, value2)`

jgiven-core/src/main/java/com/tngtech/jgiven/impl/util/ReflectionUtil.java
```diff
@@ -180,109 +164,134 @@ public static void invokeMethod( Object object, Method method, String errorDescr
...
+                    format("Not able to access field '%s'." + errorDescription, fieldName), e);
```
l-1squared, line 248: also I am not certain that passing an error description is worth it here. You are probably fine with 
```
("Not able to access field '{}' of class '{}'", fieldName, object.getClass().getSimpleName())
```

jgiven-core/src/main/java/com/tngtech/jgiven/impl/util/ReflectionUtil.java
l-1squared: Logging an error on debug level is only permissible, if you really handle the error.

l-1squared: ...But then again, if you know well how to handle an error, and you expect them to occurr, then using the exception mechnaism isn't the proper tool in the first place (at least in the Java world. Python for instance has a different view on that matter.

jgiven-core/src/main/java/com/tngtech/jgiven/base/StageNameWrapper.java
```diff
@@ -0,0 +1,36 @@
+package com.tngtech.jgiven.base;
+
+import java.util.Objects;
+
+/**
+ * Wrapper class for the name held in the stage objects.
+ * Helps when making a single stage act as the default three stages.
+ */
+public class StageNameWrapper {
```
l-1squared, line 9: Lombok would have made this sooo much easier...

l-1squared: Maybe we should introduce it, if we find the time

jgiven-core/src/main/java/com/tngtech/jgiven/base/StageNameWrapper.java
```diff
@@ -0,0 +1,36 @@
...
+    public void setStageName(String stageName) {
+        this.stageName = stageName;
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+        StageNameWrapper that = (StageNameWrapper) o;
+        return Objects.equals(stageName, that.stageName);
```
l-1squared, line 29: I think, that line might be sufficient

jgiven-core/src/main/java/com/tngtech/jgiven/impl/intercept/ByteBuddyMethodInterceptor.java
```diff
@@ -17,54 +18,71 @@
...
+    }
+
+    private void updateStageNameWrapper(Object stageObject, Method method) {
+        StageNameWrapper methodNameWrapper = getMethodNameWrapper(method);
+        if (methodNameWrapper != null) {
+            ((StageNameInternal) stageObject).__jgiven_setStageNameWrapper(methodNameWrapper);
+        }
+    }
+
+    private StageNameWrapper getMethodNameWrapper(Method method) {
```
l-1squared, line 81: using null is not as bad in private methods, but still I'd like to know if `Optional<StageNameWrapper>` would wokr here.

jgiven-tests/src/test/java/com/tngtech/jgiven/impl/ScenarioExecutorTest.java
```diff
@@ -1,88 +1,136 @@
...
+    }
 
-        public void the_next_step_is_skipped() {}
+    @Test
+    public void get_stage_name_wrapper_returns_the_last_executed_wrapper_if_given_wrapper_null() {
+        ScenarioExecutor simpleScenarioExecutorObject = new ScenarioExecutor();
+        StageNameWrapper stageNameWrapper = new StageNameWrapper("GIVEN");
 
-    }
+        doReturn(null).when((StageNameInternal) mockedStageObject).__jgiven_getStageNameWrapper();
```
l-1squared, line 95: doesn't line 95 contradict ln 93?

jgiven-tests/src/test/java/com/tngtech/jgiven/impl/ScenarioExecutorTest.java
```diff
@@ -1,88 +1,136 @@
...
 
-        public void a_NotImplementedYet_annotated_step() {}
+        assertThat(simpleScenarioExecutorObject
+                .getStageNameWrapper(mockedStageObject, null)).isSameAs(stageNameWrapper);
+    }
 
-        public void the_next_step_is_skipped() {}
+    @Test
+    public void get_stage_name_wrapper_returns_the_last_executed_wrapper_if_given_wrapper_null() {
+        ScenarioExecutor simpleScenarioExecutorObject = new ScenarioExecutor();
```
l-1squared, line 92: `scenarioExecutor` works as well, but is shorter :)

jgiven-core/src/main/java/com/tngtech/jgiven/base/StageNameWrapper.java
l-1squared: Btw: why not just `StageName`? Thats what it is supposed to represent.

jgiven-tests/src/test/java/com/tngtech/jgiven/impl/ScenarioExecutorTest.java
andru47: The stage name is not set inside mockedObject. What this test is trying to verify is that if the current object we receive as a stage doesn't have a name set to GIVEN/WHEN/THEN (thus having it null), then the last known stage we are executing is not changed. In the case of given().a_simple_test(), given() will set the current stage we are executing to GIVEN, but a_simple_test() has the internal name set to null and we don't want to change the current stage we are executing to that.

jgiven-core/src/main/java/com/tngtech/jgiven/impl/intercept/ByteBuddyMethodInterceptor.java
andru47: I think it would make things harder if we are implementing it with an Optional. If it is set to null, we deal with it in the methods from ScenarioExecutor and we are also using Objects.equals as you suggested.

jgiven-core/src/main/java/com/tngtech/jgiven/base/StageNameWrapper.java
andru47: We also need the compare of classes so we can cast the given object to a StageNameWrapper.

l-1squared:
turned out to be a bad idea, because it would have made stages would have behaved differently depending on how they were instanciated. Thus it was considered better to continue having the simple stage class behave as a single stage, despite the pointlessness of the after stage attributes for this usecase

l-1squared:
Closed because of the aforementioned issues

PR #697, opened 2021-07-26
Lombok integration

andru47:


l-1squared (approved):

jsalinaspolo:
Should we be adding Lombok as a dependency?
It is another dependency to take care of and I would argue about the value that might add. Currently, it might be working with Gradle incremental builds, but it has been problematic in the past with task avoidance. 

Not being able to have Gradle task avoidance would increase build times

l-1squared:
> Should we be adding Lombok as a dependency?
> It is another dependency to take care of and the value that might add is arguable. Currently, it might be working with Gradle incremental builds, but it has been problematic in the past with task avoidance.
> 
> Not being able to have Gradle task avoidance would increase build times

Hi,
while we are not replacing many classes, I am strongly in favor of having it. I recall several instances where I was wishing to have lombok at my hand to quickly create some classes.  Furthermore, while andru has only replaced getters, more usages should show up when we take a look at constructors. 

Considering your concerns regarding build times, I have not found any sources that corroborate that lombok messes with the task configuration avoidance. Could you elaborate on that? Furthermore, if lombok interfered with task configuration avoidance, wouldn't it then also interfere with the incremental builds?

jsalinaspolo:
Currently Gradle supports it but it was not supported until version ~4.7 because of the annotations processors.
In the past, I had issues moving to java 1.8 with lombok because there are things done through non public APIs

Personally, I do not see the benefit to use lombok because most of IDEs generates the code and by observing some codebases using Lombok, usually suffers from anemic domains models.

adrian-herscu:
Just sharing my experience after using Lombok for about 6-7 years.
For non-library projects it is great :) I really like it.

Two-three months ago I decided to package my testing framework as a set of libraries and push it to GitHub (<https://github.com/adrian-herscu/qa-automation>). BTW, it is based on JGiven.

Then, after several days I discovered the atrocities of Lombok!
I couldn't debug anymore, IntelliJ complained that the sources do not match the bytecode.

I had to configure another Maven plugin to make it de-Lombok the code (<http://anthonywhitford.com/lombok.maven/lombok-maven-plugin/delombok-mojo.html>)
and then reformat it and re-sort the imports.
Now at least I can step into my library code. But the code does not look as the original.

I think similar approach you will have to take with JGiven if you decide to use Lombok.
Otherwise your users will not be able to step into JGiven code.

l-1squared:
Thanks @adrian-herscu  for your insight on the matter. After the strong opposition of @jsalinaspolo  I refrained from going forward using lombok in our project.

Given that we do not benefit that much from lombok in thefirst place and that we would make it really hard for consumers to debug code, I think its finally time to put this idea to rest.

PR #699, opened 2021-07-27
Update github actions for test reports artifacts

andru47:


.github/workflows/jgiven_branch_and_pr_build.yml
```diff
@@ -25,3 +25,8 @@ jobs:
       - name: Run Jacoco
         if: ${{matrix.java == '8' && github.ref == 'refs/heads/master' && github.head_ref == null}}
         run: ./gradlew overallJacocoReport coveralls
+      - name: Upload test results html files
+        uses: actions/upload-artifact@v2
+        with:
+          name: test-report ${{matrix.java}} ${{matrix.junit}}
```
l-1squared, line 31: please use underscorese in the name.

.github/workflows/jgiven_branch_and_pr_build.yml
```diff
@@ -25,3 +25,8 @@ jobs:
...
+          path: /home/runner/work/JGiven/JGiven/jgiven-tests/build/reports/tests
```
l-1squared, line 32: wouldn't it make sense, to also bundle the reports of all the other artifacts?

l-1squared (approved):
good first attempt. maybe we need to refine the results later

PR #700, opened 2021-07-27
Add new guaranteed field inside a ScenarioState or a ProvidedScenarioState

### LINKED ISSUE #268: Add a guaranteed attribute to scenario states
This is similar to #255, but instead of dealing with provided states, we should have a `guaranteed` attribute on `@ScenarioState` and `@ProvidedScenarioState` which will fail with a clear exception if the stage doesn't provide this state.

andru47:
Fixes #268


jgiven-core/src/main/java/com/tngtech/jgiven/impl/StageLifecycleManager.java
```diff
@@ -66,6 +72,33 @@ private void executeLifecycleMethods(LifecycyleMethodManager<?> register, boolea
         }
     }
 
+    @SuppressWarnings("unchecked")
+    private void checkGuaranteedStatesAreInitialized() throws JGivenMissingGuaranteedScenarioStateException {
+        for (Field field: FieldCache.get(instance.getClass())
+                                    .getFieldsWithAnnotation(ProvidedScenarioState.class, ScenarioState.class)) {
+            if (field.isAnnotationPresent(ProvidedScenarioState.class)) {
+                if (field.getAnnotation(ProvidedScenarioState.class).guaranteed()) {
+                    checkInialized(field);
```
l-1squared, line 81: typo

jgiven-core/src/main/java/com/tngtech/jgiven/impl/StageLifecycleManager.java
```diff
@@ -66,6 +72,33 @@ private void executeLifecycleMethods(LifecycyleMethodManager<?> register, boolea
         }
     }
 
+    @SuppressWarnings("unchecked")
+    private void checkGuaranteedStatesAreInitialized() throws JGivenMissingGuaranteedScenarioStateException {
+        for (Field field: FieldCache.get(instance.getClass())
+                                    .getFieldsWithAnnotation(ProvidedScenarioState.class, ScenarioState.class)) {
+            if (field.isAnnotationPresent(ProvidedScenarioState.class)) {
```
l-1squared, line 79: these double if blocks can go into a separate method, such that you can reuse them

jgiven-core/src/main/java/com/tngtech/jgiven/impl/StageLifecycleManager.java
```diff
@@ -66,6 +72,33 @@ private void executeLifecycleMethods(LifecycyleMethodManager<?> register, boolea
         }
     }
 
+    @SuppressWarnings("unchecked")
+    private void checkGuaranteedStatesAreInitialized() throws JGivenMissingGuaranteedScenarioStateException {
```
l-1squared, line 76: have you thought about placing this functionality in the `ValueInjector` ? If so, why here?

jgiven-core/src/main/java/com/tngtech/jgiven/impl/StageLifecycleManager.java
```diff
@@ -44,6 +49,7 @@ boolean allAfterStageMethodsHaveBeenExecuted() {
 
     void executeAfterStageMethods(boolean fakeExecution) throws Throwable {
         executeLifecycleMethods(afterStageRegister, fakeExecution);
+        checkGuaranteedStatesAreInitialized(); //before or after the other after stage methods?
```
l-1squared, line 52: in wouldn't this functionality be more apropriate to live in the `ValueInjector`, where the actual states are read?

jgiven-core/src/main/java/com/tngtech/jgiven/impl/StageLifecycleManager.java
andru47: Not sure if we can make that work without either using instance of and hence another two if statements or method execution through resolution. You need to cast to that class (ProvidedScenarioState or ScenarioState) in order to be able to run guaranteed() on the object.

jgiven-tests/src/test/java/com/tngtech/jgiven/impl/GuaranteedStateTest.java
```diff
@@ -0,0 +1,51 @@
...
+        given().a_given_stage_with_guaranteed_field_uninitialized();
+        then().the_before_is_not_called();
+    }
+
+    @Test(expected = ClassNotFoundException.class)
+    public void assure_before_method_of_second_test_is_executed_if_guaranteed_initialized() {
+        given().a_given_stage_with_guaranteed_field_initialized();
+        then().the_before_is_not_called();
+    }
+
```
andru47, line 26: Not a fully JGiven test because I find it very hard to check the order of execution without throwing these exceptions.

l-1squared: Something like that could work. You'd have to check the `WhenTestFramework` stage for details

jgiven-core/src/main/java/com/tngtech/jgiven/impl/StageLifecycleManager.java
l-1squared: Yes, I forgot that Annotations don't have inheritance

jgiven-core/src/main/java/com/tngtech/jgiven/impl/inject/ValueInjector.java
```diff
@@ -1,139 +1,168 @@
...
+                }
+            }
+        }
+    }
+
+    private void checkInitialized(Object instance, Field field) {
+        Object value = null;
+        try {
+            value = field.get(instance);
+        } catch (IllegalAccessException e) { }
```
l-1squared, line 163: I wouldn't lump the Illegal access and the absent value excpetion together. They are two very different kinds of errors with very different reasons.

jgiven-tests/src/test/java/com/tngtech/jgiven/impl/GuaranteedStateTest.java
```diff
@@ -0,0 +1,110 @@
...
+    public void assure_before_method_of_second_test_is_executed_if_guaranteed_initialized() {
+        given().a_Jgiven_test_with_a_guaranteed_state();
+        when().the_test_is_executed();
+        then().the_report_contains_$_exception(ClassNotFoundException.class);
+    }
+
+    public static class RealTest extends ScenarioTest<RealTest.RealGiven, RealTest.RealWhen, RealTest.RealThen> {
+
+        @Test
+        @SuppressWarnings("checkstyle:MethodName")
```
l-1squared, line 36: I think a better approach to solving this is to check whether we can disable some checkstyle rules for test classes.

jgiven-tests/src/test/java/com/tngtech/jgiven/impl/GuaranteedStateTest.java
```diff
@@ -0,0 +1,110 @@
...
+        public void a_sample_test() {
+            given().a_sample_uninitialized_stage();
+            when().I_do_something();
+            then().I_did_something();
+        }
+
+        @Test
```
l-1squared, line 43: Given the errors you mentioned: Have you tried just not having the `@Test` annotation?

jgiven-tests/src/test/java/com/tngtech/jgiven/impl/GuaranteedStateTest.java
```diff
@@ -0,0 +1,110 @@
...
+                                .execute(testScenario.testClass, testScenario.testMethod).getReportModel();
+        }
+
+        @SuppressWarnings("checkstyle:MethodName")
+        public void the_report_contains_$_exception(Class<? extends Exception> givenException) {
+            assertThat(testReport.getFailedScenarios()).isNotEmpty();
+            assertThat(testReport.getFailedScenarios().get(0)
+                    .getScenarioCases().get(0).getErrorMessage()).contains(givenException.getName());
+        }
+
```
l-1squared, line 108: unnecessary emtpy line

jgiven-core/src/test/java/com/tngtech/jgiven/impl/inject/ValueInjectorTest.java
```diff
@@ -0,0 +1,43 @@
+package com.tngtech.jgiven.impl.inject;
+
+import com.tngtech.jgiven.annotation.ProvidedScenarioState;
+import com.tngtech.jgiven.annotation.ScenarioState;
+import com.tngtech.jgiven.exception.JGivenMissingGuaranteedScenarioStateException;
+import org.junit.Test;
+
+public class ValueInjectorTest {
+    private ValueInjector injector = new ValueInjector();
```
l-1squared, line 9: Btw. Is there really no Unit test for value injectors yet?  
It might be a good idea to use the opportunity to also assert the behavior of the required attribute / the injection mechanism in general.

andru47: Added some tests for the required field in the fixing commit.

jgiven-core/src/main/java/com/tngtech/jgiven/impl/StageLifecycleManager.java
```diff
@@ -4,10 +4,15 @@
 import com.tngtech.jgiven.annotation.AfterStage;
 import com.tngtech.jgiven.annotation.BeforeScenario;
 import com.tngtech.jgiven.annotation.BeforeStage;
+import com.tngtech.jgiven.annotation.ProvidedScenarioState;
```
l-1squared, line 7: these should not be here.

jgiven-core/src/test/java/com/tngtech/jgiven/impl/inject/ValueInjectorTest.java
l-1squared: Bonus, have one of the "does not throw anything" tests assert that you actually can inject & read values. (for instance, by using the injector to initialize a public field)

jgiven-tests/src/main/java/com/tngtech/jgiven/tests/GuaranteedFieldRealTest.java
```diff
@@ -0,0 +1,51 @@
+package com.tngtech.jgiven.tests;
+
+import com.tngtech.jgiven.Stage;
+import com.tngtech.jgiven.annotation.BeforeStage;
+import com.tngtech.jgiven.annotation.ProvidedScenarioState;
+import com.tngtech.jgiven.junit.ScenarioTest;
+import org.junit.Test;
+
+public class GuaranteedFieldRealTest extends ScenarioTest<GuaranteedFieldRealTest.RealGiven, GuaranteedFieldRealTest.RealWhen, GuaranteedFieldRealTest.RealThen> {
```
l-1squared, line 9: +1 for that name :)

jgiven-tests/src/test/java/com/tngtech/jgiven/impl/GuaranteedStateTest.java
```diff
@@ -0,0 +1,54 @@
...
+        public void a_Jgiven_test_with_a_guaranteed_null_state() {
+            testScenario = new TestScenarioRepository.TestScenario(GuaranteedFieldRealTest.class, "a_sample_test");
+        }
+
+        public void a_Jgiven_test_with_a_guaranteed_state() {
+            testScenario = new TestScenarioRepository.TestScenario(GuaranteedFieldRealTest.class, "a_sample_initialized_test");
+        }
+
+        public void the_test_is_executed() {
+            testReport = TestExecutor.getExecutor(TestFramework.JUnit)
```
l-1squared, line 44: technically fine, but I think it would be nicer to read, if you separated that into two statements. As a side effect, if one of those methods throws an unexpected exception, it will be easier to figure out which of the two threw it. 
(its the avoid train-wreck rule :))

l-1squared (approved):

l-1squared (approved):
You could squash & rebase your commits.

PR #778, opened 2021-12-08
Bugfix/issue 755

l-1squared:


jgiven-core/src/main/java/com/tngtech/jgiven/impl/ScenarioModelBuilder.java
```diff
@@ -293,18 +328,25 @@ public void stepMethodFinished(long durationInNanos, boolean hasNestedSteps) {
...
                 parentSteps.pop();
+                discrepancyOnLayer.pop();
             }
         }
 
         if (!hasNestedSteps && !parentSteps.isEmpty()) {
             currentStep = parentSteps.peek();
         }
+
+        if (discrepancyOnLayer.peek() > 0) {
```
l-1squared, line 347: the condition should be placed in the decrement discrepancy method

fudler (approved):

PR #779, opened 2021-12-09
Increase growth cutoff for ProxyClassPerformanceTest

l-1squared:
Previously the test failed for a scenario where the reported memory usage would be constant but increased for the last recorded item. In such a case the reported growth would not be 0, but looking at the record would reveal no continuous growth.

Therefore we now allow the last item in the record to be an increment by 1.


jgiven-core/src/test/java/com/tngtech/jgiven/impl/ProxyClassPerformanceTest.java
```diff
@@ -14,6 +14,8 @@
 
     private static final int NUMBER_OF_RUNS = 10000;
     private static final int PROBE_INTERVAL = 100;
+    private static final double GROWTH_EXPECTED_IF_LAST_REPORTED_GROWTH_IS_POSITIVE = 0.11;
+
```
fudler, line 18: This should rather be 0.011 :)

fudler (approved):

PR #783, opened 2021-12-16
Feature/automated release

l-1squared:


l-1squared:
@NilsOliverLinden  ich hab dich mal hinzugefügt, weil ich das script während des releasprozesses aufrufen möchte

scripts/source_files/helper_functions.sh
```diff
@@ -1,34 +1,28 @@
...
-function update_version(){
-  [ $# -eq 2 ] || return 11
+function __update_a_version_property(){
+  [ $# -eq 3 ] || return 11
   local target_version="$1"
   local target_file="$2"
-  local version_property="version"
-  local version_matcher="${version_property}=.*"
+  local version_property="$3"
+  local version_matcher="${version_property} ?= ?.*"
```
fudler, line 8: For reference, what does ?= do here?

l-1squared: that is a regex expression. so in the current version we have ( ?= ?). the parentheses denote a capturing group that allows us to record the contents and drop them into our replacement expression later. 
The ? is an "optional" signifier, denoting that the character (whitespace in our case) need not be there.

.github/workflows/test-bash-functions.yml
```diff
@@ -8,11 +8,11 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - name: Run release functions test
-        run: "bash scripts/test_release_functions.sh"
+        run: "bash scripts/tests/test_release_functions.sh"
```
NilsOliverLinden, line 11: I do not know if that is a problem. But locally I can not execute those shell scripts located in tests. 

The same is true for the referenced files but this may be fine as you execute them from within other scripts? IDK.

.gitattributes
```diff
@@ -0,0 +1,2 @@
+# Declare files that will always have LF line endings on checkout.
+*.sh text eol=lf
```
NilsOliverLinden, line 2: Can you enlighten me why you need this?
Never have seen that option within one of my projects.

scripts/publishReleaseOnGithub.sh
```diff
@@ -0,0 +1,31 @@
...
+    'name' : 'Release ${release_tag}',
+    'body' : 'Version ${version} release of JGiven.\nSource code may be found in Maven Central: https://search.maven.org/search?q=com.tngtech.jgiven',
+    'draft' : ${is_draft}
+  }"
+  replace_single_by_double_quotes "${json_adjacent_body}"
+  return $?
+}
+
+body=$(create_message_body "$1") || { code=$?;  echo "Failed to create message body with error ${code}"; exit ${code}; }
+action_token=$([ -n "$2" ] && echo "$2") || { echo "Failed to parse action token"; exit 1; }
```
NilsOliverLinden, line 26: The script is hard to understand as I do not know what the second parameter is that is supposed to be handed to the script.

I think a better way of parsing input parsing ist something like this

while [[ -n "${1-}" ]] ; do
  case "${1}" in
    --old-version*)
      OLD_VERSION=${1#*=}
      ;;
    --new-version*)
      NEW_VERSION=${1#*=}
      ;;
    --repository-username*)
      SONATYPE_USERNAME=${1#*=}
      ;;
    --repository-password*)
      SONATYPE_PASSWORD=${1#*=}
      ;;
    --signing-key*)
      GPG_SIGNING_KEY=${1#*=}
      ;;
    --signing-password*)
      GPG_SIGNING_PASSWORD=${1#*=}
      ;;
    *)
      echo "Unknown option '${1}'"
      exit 1
      ;;
  esac
  shift
done

Have you considered something like that?

scripts/publishReleaseOnGithub.sh
```diff
@@ -0,0 +1,31 @@
...
+curl -X POST \
+  -H "Accept: application/vnd.github.v3+json" \
+  https://api.github.com/repos/TNG/JGiven/releases \
```
NilsOliverLinden, line 29: I think this means that your user must have write access to the JGiven repository. have you ensured/checked this?

NilsOliverLinden (commented):
Thanks for adding me. ;) I can not say much about your testing scripts. Have never done shell script testing. I focused more on the release scripts. :)

.github/workflows/test-bash-functions.yml
l-1squared: Weird, I can execute them directly... 
What was the error?

.gitattributes
l-1squared: I work on windows and so usually my shell scripts get checked out with crlf. This makes it impossible to run them via wsl.

scripts/publishReleaseOnGithub.sh
l-1squared: Yes I need write access to the repo. But my JGiven token should have that

scripts/publishReleaseOnGithub.sh
l-1squared: Hmm, input processing based on options... Why not.

.github/workflows/test-bash-functions.yml
NilsOliverLinden: No error at all. The files are listet with ls -a as not executable. Maybe just a local problem.. Weird.

NilsOliverLinden (dismissed):

fudler (dismissed):

CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+# v1.2
+## New features
+* JGiven now supports Spock 2 [#784](https://github.com/TNG/JGiven/pull/784) (thanks to jsalinospolo)
```
jsalinaspolo, line 3: Typo in the username.

@jsalinaspolo

l-1squared: Sorry for that

scripts/source_files/helper_functions.sh
l-1squared: Had to get rid of the regex, regex are evaluated differently on MacOS (for our android tests) and Unix (all the rest). it might have been possible but it would have been confusing at least.

fudler (approved):

PR #784, opened 2021-12-16
Add support for spock2

jsalinaspolo:


jgiven-spock/src/test/groovy/com/tngtech/jgiven/spock/stages/When.groovy
```diff
@@ -7,4 +7,4 @@ class When extends Stage<When> {
     When some_action() {
         self()
     }
-}
\ No newline at end of file
+}
```
l-1squared, line 10: Thanks for polishing :)

jgiven-spock2/src/main/groovy/com/tngtech/jgiven/spock/ScenarioSpec.groovy
```diff
@@ -0,0 +1,58 @@
...
+import net.bytebuddy.description.annotation.AnnotationDescription
+import net.bytebuddy.dynamic.loading.ClassLoadingStrategy
+import net.bytebuddy.implementation.SuperMethodCall
+import net.bytebuddy.matcher.ElementMatchers
+import org.junit.ClassRule
+import org.junit.Rule
+import spock.lang.Shared
+import spock.lang.Specification
+
+class ScenarioSpec<GIVEN, WHEN, THEN> extends Specification {
```
l-1squared, line 18: Maybe add javadoc and an `@since` annotation here aswell? After all, the entire module will only be available from 1.2.0 onwards

jgiven-spock2/build.gradle
```diff
@@ -0,0 +1,27 @@
...
+    spock2_version = '2.0-groovy-3.0'
+    groovy_version = '3.0.9'
+}
+
+dependencies {
+    api project(':jgiven-junit')
+
+    implementation "org.codehaus.groovy:groovy:$groovy_version"
+    implementation "org.spockframework:spock-core:$spock2_version"
+    implementation "org.spockframework:spock-junit4:$spock2_version"
```
l-1squared, line 19: Question: wouldn't it make sense to try to align the spock2 package with JUnit 5 directly? If I recall correctly Spock2 is JUnit5 based and upgrading the published jgiven package later should be significantly harder.

l-1squared (commented):
Thanks a lot for taking the time to do this. I am excited.  I'd like to discuss some remarks though before merging.

jgiven-spock2/build.gradle
jsalinaspolo: Spock2 uses JUnit Platform used in JUnit5, but it does not use JUnit5. Therefore can not use the Junit5 Extensions.

At the end, it is an implementation detail for the user of the library. The user will not see that internally we are using JUnit4 Rules, but we can migrate to native Spock at some point in the future.

Using Spock2 allows users of the library to migrate to JDK17

jgiven-spock2/src/main/groovy/com/tngtech/jgiven/spock/ScenarioSpec.groovy
```diff
@@ -0,0 +1,58 @@
...
+
+    Scenario<GIVEN, WHEN, THEN> createScenario() {
+        Class<GIVEN> givenClass = addDoNotInterceptToMetaClass(new TypeToken<GIVEN>(getClass()) {}.getRawType())
+        Class<WHEN> whenClass = addDoNotInterceptToMetaClass(new TypeToken<WHEN>(getClass()) {}.getRawType())
+        Class<THEN> thenClass = addDoNotInterceptToMetaClass(new TypeToken<THEN>(getClass()) {}.getRawType())
+
+        new Scenario<GIVEN, WHEN, THEN>(givenClass, whenClass, thenClass)
+    }
+
+    private <T> Class<T> addDoNotInterceptToMetaClass(Class<T> clazz) {
```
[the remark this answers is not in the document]
jsalinaspolo: I think it is related with Groovy 3 vs Groovy 2
The Generated class of MetaClass comes as a method to intercept

jgiven-spock2/.gitignore
```diff
@@ -0,0 +1 @@
+/target
```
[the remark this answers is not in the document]
jsalinaspolo: copy/paste from spock1 module, that was copied from jgiven-junit 😆 

Fixing it

l-1squared (approved):

PR #807, opened 2022-01-17
fixes #806: add a default definition for JUnit 5 @Tag annotations onl…

### LINKED ISSUE #806: JUnit 5 tags are not customizable
JUnit 5 `@Tag` annotations are automatically configured as JGiven tags in JGivenExtension. However, this automatic configuration cannot be overriden or customized.

I would like to be able to change the colour, the name and the description of these tags. I tried using a custom configuration for JUnit 5 `@Tag` annotation by using a subclass of AbstractJGivenConfiguration, but my custom definition gets overriden by the default orange one which I don't like.

I think JGivenExtension should define a default configuration for JUnit 5 `@Tag` annotation only if one has not been provided by a custom configuration.

sebphil:
…y if a custom definition has not been provided by a custom configuration implementation

### DIFF
--- a/jgiven-junit5/src/main/java/com/tngtech/jgiven/junit5/JGivenExtension.java
+++ b/jgiven-junit5/src/main/java/com/tngtech/jgiven/junit5/JGivenExtension.java
@@ -11,6 +11,7 @@ import org.junit.jupiter.api.extension.*;
 import org.junit.jupiter.api.extension.ExtensionContext.Namespace;
 
 import com.tngtech.jgiven.base.ScenarioTestBase;
+import com.tngtech.jgiven.config.AbstractJGivenConfiguration;
 import com.tngtech.jgiven.config.ConfigurationUtil;
 import com.tngtech.jgiven.impl.ScenarioBase;
 import com.tngtech.jgiven.impl.ScenarioHolder;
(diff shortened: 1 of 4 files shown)


l-1squared:
Hi,
thanks for stepping up and going for fixing the issue directly. The change you made looks just fine, but I would be happy with a test for the new behavior.

sebphil:
I did not find a test for the previous behavior, so I'm unsure where and how to test this. Perhaps I should add a new test case into JUnit5ExtensionTest ? Or maybe create a new test class, since I would have to add a custom configuration using `@JgivenConfiguration` on the test class. I'm guessing I should add 2 tests, one with a custom configuration, and one without. What do you think ?

l-1squared:
> I did not find a test for the previous behavior, so I'm unsure where and how to test this. Perhaps I should add a new test case into JUnit5ExtensionTest ? Or maybe create a new test class, since I would have to add a custom configuration using `@JgivenConfiguration` on the test class. I'm guessing I should add 2 tests, one with a custom configuration, and one without. What do you think ?

Sorry for not answering you sooner. If there are no test cases, then two tests in the junit5 package would be ideal.

l-1squared (approved):

PR #811, opened 2022-01-19
Feature/issue 806 cleanup

l-1squared:


jgiven-junit5/src/main/java/com/tngtech/jgiven/junit5/JGivenExtension.java
```diff
@@ -35,61 +35,62 @@
...
+    AfterEachCallback {
 
-    private static final Namespace NAMESPACE = Namespace.create( "com.tngtech.jgiven" );
+    private static final Namespace NAMESPACE = Namespace.create("com.tngtech.jgiven");
 
     private static final String REPORT_MODEL = "report-model";
 
     @Override
-    public void beforeAll( ExtensionContext context ) throws Exception {
+    public void beforeAll(ExtensionContext context) {
```
fudler, line 49: So there was never an exception...?

jgiven-junit5/src/main/java/com/tngtech/jgiven/junit5/JGivenExtension.java
```diff
@@ -35,61 +35,62 @@
...
-        getScenario().startScenario( context.getTestClass().get(), context.getTestMethod().get(),
-            ArgumentReflectionUtil.getNamedArgs( context ) );
+    public void beforeEach(ExtensionContext context) {
+        getScenario().startScenario(context.getTestClass().get(), context.getTestMethod().get(),
+            ArgumentReflectionUtil.getNamedArgs(context));
     }
 
     @Override
-    public void afterEach( ExtensionContext context ) throws Exception {
+    public void afterEach(ExtensionContext context) throws Exception {
```
fudler, line 78: Is this throwing of an exception then still necessary? Or would it lead to more refactoring?

jgiven-junit5/src/main/java/com/tngtech/jgiven/junit5/JGivenExtension.java
l-1squared: No that came in, because the interface Method declares an Exception. Which it needs to, because there could be an implementation somewhere, where a checked Exception is actually produced.

jgiven-junit5/src/main/java/com/tngtech/jgiven/junit5/JGivenExtension.java
l-1squared: Well we throw an Exception in ln 91. So we need to declare that we do.

fudler (approved):

Issue #823, opened 2022-02-01
Unable to run tests in parallel with maven plugin 


Hatzen:
I tried to run our Maven builds with multiple threads in order to make them faster. Maven complains about the jgiven plugin because it is not marked thread-safe.
For example:
```
mvn -T 1.5C clean install 
```
leads to following error without the test getting run.
```
....
[WARNING] *****************************************************************
[WARNING] * Your build is requesting parallel execution, but project      *
[WARNING] * contains the following plugin(s) that have goals not marked   *
[WARNING] * as @threadSafe to support parallel building.                  *
[WARNING] * While this /may/ work fine, please look for plugin updates    *
[WARNING] * and/or request plugins be made thread-safe.                   *
[WARNING] * If reporting an issue, report it against the plugin in        *
[WARNING] * question, not against maven-core                              *
[WARNING] *****************************************************************
[WARNING] The following plugins are not marked @threadSafe in SOME_PROJECT_NAME:
[WARNING] com.tngtech.jgiven:jgiven-maven-plugin:1.1.0
[WARNING] Enable debug to see more precisely which goals are not marked @threadSafe.
[WARNING] *****************************************************************
...
```  
After a short research it should be quite simple like adding `threadSafe=true` to the Mojo Annotation.  
https://github.com/TNG/JGiven/blob/37d2ee19c2724136e8eeb3ceb69a6ec43b16789b/jgiven-maven-plugin/src/main/java/com/tngtech/jgiven/maven/JGivenReportMojo.java#L21
But i am not very familiar with the code base and if it is missing on purpose? Maybe creating the test reports or something like this could create problems?  
But as I cannot find any issues except some different parallel execution errors i would ask to give it a trial to speed up our very large test base.

l-1squared:
I will take care of the PR tomorrow, I want to take some time to run (and write) a test to see that this really fixes the parallel exectuion issue, before announcing that it does

l-1squared:
From a cursory look, I would say that the Thread safe tag is appropriate. However, I want to take this as an apportunity to look at a parallelization issue that cropped up when using TestNG. 

Regardless of that, if I read the internet correctly you are trying to have maven execute module builds in parallel, and I would expect those to be fine.

PR #824, opened 2022-02-01
Annotate maven plugins mojo with threadSafe true

Hatzen:
for running maven builds in parallel.  Regarding to https://github.com/TNG/JGiven/issues/823

Solution gained from:  
https://cwiki.apache.org/confluence/display/MAVEN/Parallel+builds+in+Maven+3  
and   
https://stackoverflow.com/questions/24448623/how-to-mark-maven-plugin-threadsafe


l-1squared:
Hi @Hatzen, thanks for tackling the issue you found this morning. 
Please note that you need to sign your commit in order for us to place it in master.

Hatzen:
Thanks for advice, i thought editing with github should sign correctly. I just manipulated the commit now it should be fine.

l-1squared (approved):

PR #872, opened 2022-04-13
Issue-868: make sure that the parents of a tag are known to the repor…

l-1squared:
…t model

Commit fd978254a2494e8dde4bc194ea42251647c8fef2 in this repo removed a side effect during tag creation that ensured that parent tags of a declared tag were declared in the tag map of the report model. This is a prerequisite for the html app to work. Unfortuneately there was no test that ensured that the report model got the info.

Here, we changed the model such that the Tag creator actually publishes the parents that a tag has, such that the report model may consume them.

### DIFF
--- a/jgiven-core/src/main/java/com/tngtech/jgiven/impl/ScenarioModelBuilder.java
+++ b/jgiven-core/src/main/java/com/tngtech/jgiven/impl/ScenarioModelBuilder.java
@@ -22,6 +22,7 @@ import com.tngtech.jgiven.format.ObjectFormatter;
 import com.tngtech.jgiven.impl.format.ParameterFormattingUtil;
 import com.tngtech.jgiven.impl.intercept.ScenarioListener;
 import com.tngtech.jgiven.impl.params.DefaultAsProvider;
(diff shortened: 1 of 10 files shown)


jgiven-core/src/main/java/com/tngtech/jgiven/impl/ScenarioModelBuilder.java
```diff
@@ -496,17 +496,19 @@ private void addTags(Annotation... annotations) {
...
+    private void addTags(ResolvedTags tags) {
         if (tags.isEmpty()) {
             return;
         }
 
         if (reportModel != null) {
-            this.reportModel.addTags(tags);
+            this.reportModel.addTags(tags.getDeclaredTags());
+            //The report model needs to declare the parent tags in a tag map, or the tags cannot be displayed.
+            this.reportModel.addTags(tags.getParents());
```
l-1squared, line 507: maybe we could tryto pass on resolved tags to the report model and the scenario model...

fudler (approved):

PR #882, opened 2022-05-11
Fix/release 1.2.1

l-1squared:


example-projects/kotlin/build.gradle
```diff
@@ -1,6 +1,6 @@
 plugins {
```
l-1squared, line 1: check if one can use a buildscript rather than this plugin version mashing.

PR #942, opened 2022-08-26
Fix/restructure dependency handling

l-1squared:


.github/workflows/run_renovate.yml
```diff
@@ -0,0 +1,20 @@
+name: Renovate-Dry-Run
```
l-1squared, line 1: this file needs to go

renovate.json
```diff
@@ -1,9 +1,13 @@
 {
+  "dryRun": "full",
+  "baseBranches": [ "fix/restructure-dependency-handling" ],
+  "repositories": ["TNG/JGiven"],
```
l-1squared, line 4: this needs to go

l-1squared:
TODO: check that version override actually works given that the junit version might be drawn from the junit-bom

.github/workflows/jgiven_android_build.yml
```diff
@@ -1,5 +1,5 @@
 name: "Build and test android plugin and projects"
-on: [push, pull_request]
+on: [pull_request]
```
fudler, line 2: Is this a huge improvement to remove the push?

gradle/libs.versions.toml
```diff
@@ -0,0 +1,35 @@
+[versions]
+aspectj-version = "1.9.9.1"
+byteBuddy-version = "1.12.8"
+junit4-version = "4.13.2"
+slf4j-version = "1.7.35"
+spring-version = "5.3.19"
+
+[libraries]
+assertj = { module = "org.assertj:assertj-core", version = "3.22.0" }
```
fudler, line 9: Could it be of advantage to move all version numbers to the [versions]-part?

jgiven-spring-junit5/build.gradle
```diff
@@ -7,20 +7,19 @@ description = "Module for using Spring dependency injection together with JGiven
...
-    testImplementation "org.springframework:spring-tx:$springVersion"
-    testImplementation "org.springframework:spring-context:$springVersion"
-    testImplementation "org.springframework:spring-test:$springVersion"
-    testImplementation "org.springframework:spring-jdbc:$springVersion"
-    testImplementation "org.hsqldb:hsqldb:$hsqldbVersion"
-    testImplementation "org.aspectj:aspectjrt:1.9.9.1"
-    testImplementation "org.aspectj:aspectjweaver:1.9.9.1"
+    testImplementation 'org.junit.jupiter:junit-jupiter-engine'
+    testImplementation 'org.junit.platform:junit-platform-runner'
+    testImplementation 'org.junit.jupiter:junit-jupiter-params'
```
fudler, line 19: Couldn't we also bundle these three dependencies? I mean, we need them at several places.

gradle/libs.versions.toml
l-1squared: I would only move them there if we intend to reference or reuse them. Otherwise, you'll get a ton of versions that are only used once and I feel that that just clutters files

.github/workflows/jgiven_android_build.yml
l-1squared: Goot thing we're talking about this. That was only to avoid excessive builds during testing how the change affects renovate (Because with this change we should get better coverage of renovate for our dependencies)

jgiven-spring-junit5/build.gradle
l-1squared: A point of contention.
We should discuss this in person

.github/workflows/run_renovate.yml
l-1squared: I kept a chaned file in the repo. Should be nice-to have, if we ever need to update the renovate.json

.github/workflows/jgiven_android_build.yml
fudler: Then this should be enabled again in the final commit.

jgiven-spring-junit5/build.gradle
fudler: That's maybe still a point of discussion. In my opinion this could "purge" the code a bit.

settings.gradle
```diff
@@ -20,6 +20,18 @@ plugins {
     id 'com.gradle.enterprise' version '3.11.1'
 }
 
+dependencyResolutionManagement {
+    versionCatalogs {
+        junitVariableVersionLibs {
+            from(files('gradle/libs.versions.toml'))
+            def junitVersion = rootProject.hasProperty('junitVersion') ? rootProject.junitVersion : null
```
fudler, line 27: Isn't this the replacement of build.gradle l.60? Why was null chosen here?

l-1squared: well, we only overwrite the original if it is not null. But you are right, this can be written more elegantly.

fudler (approved):

PR #952, opened 2022-09-09
Issue-947: Fix cast failure

l-1squared:
For an unknown reason the use of the google functional api produces an error if openapi is present on the classpath. Just migrating this single-use lambda to the java functional api should solve the issue


fudler (approved):

l-1squared:
PR has already been approved and no meaningful changes were made since then

PR #953, opened 2022-09-15
Checkout actions fetches whole repository

l-1squared:
Previously the repository did not checkout a single branch but actually a single commit, which it made into a unique ref for its own repo. Consequently, when attempting to rebase the entire history of gh-pages onto that commit the rebase would fail, because the base in the repo wasn't the actual master branch but only its latest commit which resulted in an unfitting history of that commit and a failing rebase.

Now that we check out everything from the start, this should work fine.


l-1squared:
Force merged. Change is comparatively simple and has been tested. Also I want it this to work for the next release, which is due

PR #1018, opened 2022-10-21
Issue-802: Ensure commit is on graph

l-1squared:
If the bot does not have admin rights on github, the push to master will fail. However, the creation of the tag will be successful creating a tag object that is not on the graph. To avoid this we place the commit on a branch an push that branch, to be merged manually after the release

### DIFF
--- a/scripts/source_files/release_functions.sh
+++ b/scripts/source_files/release_functions.sh
@@ -1,29 +1,28 @@
 #!/usr/bin/env bash
 
-function find_gradle_property(){
-    [[ "$(declare -p $1 2>/dev/null)" =~ "declare -a" ]] || return 11
-    local -n output_target=$1
-    for command_line_argument in "$@"; do
-        if [[ "${command_line_argument}" =~ ^-P.*=.* ]];then
-          output_target+=("${command_line_argument}")
-        fi
-    done
+function find_gradle_property() {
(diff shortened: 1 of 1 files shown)


scripts/source_files/release_functions.sh
```diff
@@ -1,17 +1,28 @@
...
+  done
 }
 
-function releaseRepositoryAndPushVersion()
-{
+function push_version() {
+  echo Pushing version and tag to GitHub repository...
+  git push && push_result="success" || push_result="failure"
+  if [ push_result != "success" ]; then
+    branch_name="release/${VERSION_PREFIXED}"
```
fudler, line 17: Maybe it's too early in the morning, but where is the VERSION_PREFIXED defined? 
Further, at the current point the clean-up of this branch has to be done manually, doesn't it?

fudler: Ah, and this branch also has to be provided to other scripts somehow, I guess?

l-1squared: VERSION_PREFIXED is defined in the release.sh. Not the best design, I know...

l-1squared: > Ah, and this branch also has to be provided to other scripts somehow, I guess?

Good catch. I have not thought of that!

l-1squared:
Since this solution is fragile and only mitigates the problem (we still have to merge manuall) we'll keep handing Admin rights to the release bot part of the process.

PR #1157, opened 2023-01-16
Fix broken build shield

l-1squared:
Route of build status had to be changed, because the build provider identified the need for a breaking change in its service. Details see here: https://github.com/badges/shields/issues/8671


l-1squared:
Really to trivial and none of the tests would detect if the change was defective anyway

PR #1158, opened 2023-01-16
Increase minimum version to Java 11

l-1squared:
Java 11 is the baseline version for most current version of Jgivens dependencies (with the notable exception of spring). Thus in order to keep them up to date we need to also update the minimum supported version.

With the version change having dedicated java8 and java9 test modules has lost its purpose, so the test whether Jgiven can work with Lambda generating steps has been folded into the general JGiven-Tests module.


l-1squared:
TODO: Change required checks to fit the new names.

build.gradle
```diff
@@ -13,7 +13,7 @@ plugins {
 apply plugin: 'project-report'
 
 description = 'JGiven - BDD in plain Java'
-def jacocoEnabled = !JavaVersion.current().isJava11Compatible()
+def jacocoEnabled = !JavaVersion.current().isJava12Compatible()
```
fudler, line 16: Why do we change from 11 to 12 at this place?

l-1squared: Simple, we don't need Jacoco to run for all versions. Previously, it ran only for Java8, so due to bumping up to 11 we now only run it for Java 11.

fudler: Sorry, so you meant we only need to run it for Java 12?

l-1squared: Not exactly, we run it for Java 11 and omit it for all versions above that

fudler: :facepalm:  Yeah, I see. Thanks for the clarification!

fudler (approved):

l-1squared:
Power merge because:
* most relevant changes were already reviewed and accepted

PR #1205, opened 2023-02-22
Feature/gradle 8 preparation

l-1squared:
Removes gradle incompatibility warnings

### DIFF
--- a/jgiven-maven-plugin/build.gradle
+++ b/jgiven-maven-plugin/build.gradle
@@ -24,7 +24,7 @@ task generatePom(type: Copy, dependsOn: copyClasses) {
 }
 
 task generateMavenPlugin(type: CrossPlatformExec, dependsOn: generatePom) {
-    // currenlty it seems to be the more or less only clean solution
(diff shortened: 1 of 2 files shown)


jgiven-maven-plugin/build.gradle
```diff
@@ -43,14 +43,20 @@ generateMavenPlugin.configure {
...
         include '**/*.xml'
+        exclude '**/*.class'
     }
     into 'build/classes/java/main'
+    javadoc.dependsOn += it
+    checkstyleMain.mustRunAfter(it)
 }
 
-task copyPom(type: Copy, dependsOn: generatePom) {
+tasks.register("copyPom", Copy){it
```
fudler, line 58: it -> it.dependsOn ..?

l-1squared: indeed...

l-1squared: I guess this only works because groovy uses `it` as default argument, and we technically didn't specify one... .Ahh groovy

fudler (approved):

PR #1210, opened 2023-02-24
Fix version of github actions

l-1squared:
This hopefully improves the chance that the CI still works for older commits when they need to be reviewed.


fudler (approved):

fudler (commented):
Looks fine, but this one test doesn't ^^

l-1squared:
> Looks fine, but this one test doesn't ^^

Luckily we can mostly ignore the test. Its an issue with the android emulator

PR #1236, opened 2023-03-20
Ensure the HTML App test actually has a browser to run on.

l-1squared:


.github/workflows/jgiven_branch_and_pr_build.yml
```diff
@@ -12,6 +12,8 @@ jobs:
         java: [ '11', '13']
         junit: ['4.12','4.13']
     steps:
+      - name: Ensure firefox is installed
+        run: sudo apt-get install firefox
```
fudler, line 16: Will the sudo step work or could it crash due to the lack of credentials/permissions?

.github/workflows/jgiven_examples_build.yml
```diff
@@ -22,7 +22,8 @@ jobs:
           && runGradleTestOnGivenProject example-projects/kotlin/build.gradle 1.1-t
       - name: Test JUnit5 Example Project
         run: >
-          source scripts/source_files/helper_functions.sh
+          sudo apt-get install firefox
+          && source scripts/source_files/helper_functions.sh
```
fudler, line 26: Same as above.

jgiven-tests/src/test/java/com/tngtech/jgiven/report/html5/Html5AppTest.java
```diff
@@ -4,37 +4,21 @@
...
 import org.junit.runner.RunWith;
-import org.openqa.selenium.Dimension;
 import org.openqa.selenium.OutputType;
 import org.openqa.selenium.TakesScreenshot;
 import org.openqa.selenium.WebDriver;
-import org.openqa.selenium.chrome.ChromeDriver;
-import org.openqa.selenium.chrome.ChromeOptions;
+
+import java.io.IOException;
+import java.util.*;
```
fudler, line 21: Do we need this huge import?

.github/workflows/jgiven_branch_and_pr_build.yml
l-1squared: we generally have root and installation permissions on the ci runners. (We also install jeckyll for releases and sbt stuff for our tests).
In fact, it already worked: https://github.com/TNG/JGiven/actions/runs/4465801439/jobs/7843220117

jgiven-tests/src/test/java/com/tngtech/jgiven/report/html5/Html5AppTest.java
l-1squared: probably not

.github/workflows/jgiven_branch_and_pr_build.yml
fudler: Then everything is fine ;)

Issue #54, opened 2015-01-11
Create a AsciiDoc Reporter


janschaefer:
Having a reporter that generates AsciiDoc would have several benefits:
* The result can be transformed into many other formats like HTML or PDF
* The resulting files could be included in an overall AsciiDoc documentation that is completely separate of JGIven.

janschaefer:
Scenarios: http://jgiven.org/jgiven-report/html5/#/tag/Feature/AsciiDoc%20Report

ahus1:
I see that the AsciiDocReportModelVisitor implements ReportModelHandler, while a lot of others implement ReportModelVisitor that can access more details from the report. Can you please share your thoughts on how this should work?

How would I acccess information from tags and attachments and put them into the asciidoc report?

janschaefer:
The AsciiDoc reporter is really alpha state. If you want to work on that you are free to change the implementation however you want :-)

Airblader:
I think the best way to go here is to first come up with an example asciidoc document to get a feeling for what generated documents should look like. That's what I'll start with, anyway. :-)

A first few tests show me, however, that we will need to deliver some theme / style file. At least for reference. For example, for formatting reasons I'd use a table for displaying a scenario, but inside the asciidoc document there's no way to turn off parity-dependent backgrounds (odd / even rows). :-( Using a plain paragraph just doesn't look all that good.

rdmueller:
cool feature. Is there still someone working on it?
Is there somewhere a demo of a report which shows what already has been done?

janschaefer:
No, nobody is working on this. There is an early state that implements some basic features (`AsciiDocReportGenerator`). There is a also a rudimentary test that it works (`AsciiDocReportGeneratorTest`)

hvennekate:
Hi @johthor !
I'm going to extract the ascii doc report generator and would otherwise like to incorporate your commits so far. Is that fine with you?

johthor:
@hvennekate i took some closer look at the current state of the code. 

All code that is currently present should work as intended. But it isn't integrated with the current changes on `main` branch.

The thing that is still missing is a kind of hierarchical tag tree, like the one in the left sidebar of the HTML reporter.

PR #1299, opened 2023-05-24
Improve the AsciiDoc reporter

johthor:
Improve the AsciiDoc reporter

Some time ago in one of our projects we introduced documentation via AsciiDoc and later also some JGiven based behavior tests. Unfortunately the interactive HTML report was not compatible with the AsciiDoc based documentation.

When we found the AsciiDoc reporter via #54 we had high hopes to be able to include the produced AsciiDoc report into our documentation. Only to realize that data tables as step parameters where not supported while most of our scenarios
relied heavily on those data tables.

So some month ago we started with improving the AsciiDoc reporter by adding support for all the features we needed. Now we have reached a development stage where most of our needs are met but we don't know how to continue and if the AsciiDoc reporter is even still a wanted feature.

An example for the generated AsciiDoc files is pushed to https://github.com/eCubeGmbH/JGiven/tree/asciiDocReport under `current` while `goal` contains some ideas what could be possible.

As discussed in #54 we removed the `ReportModelHandler` again and reverted to a `ReportModelVisitor` based approach. Where the `AsciiDocReportBlockConverter` is reponsible for transforming data from the model into AsciiDoc text blocks that are later only printed into the corresponding files.

The following features and properties are supported for now:

- overall report
  - [x] statistics summary table
- features
  - [x] class name / name (optional)
  - [x] description (optional)
  - [x] mini statistic
- scenarios
  - [x] description
  - [x] extended description (optional)
  - [x] status
  - [x] duration
  - [x] tags (optional)
- standalone scenario cases
  - [x] case number
  - [x] description (optional)
  - [x] status icon
  - [x] duration
  - [x] error message (optional)
  - [x] stack trace (optional)
- scenario cases as table
  - [x] case number
  - [x] parameters
  - [x] description (optional)
  - [x] status icon
  - [x] duration
  - [x] error message (optional)
  - [x] stack trace (optional)
- steps
  - [x] status
  - [x] duration
  - [x] description
  - [x] section titles
  - [x] nested steps
  - [ ] attachments _not to be added in this PR, it is already quite big_


l-1squared:
Great! :)
Many thanks in advance for that contribution! I will have a look at it as soon as I find the time. (Probably next week)

jgiven-core/src/main/java/com/tngtech/jgiven/report/asciidoc/CasesTableImpl.java
```diff
@@ -0,0 +1,74 @@
...
+        List<ScenarioCaseModel> scenarioCases = scenarioModel.getScenarioCases();
+        if (!scenarioCases.isEmpty() && scenarioCases.get(0).hasDescription()) {
+            placeHoldersList.add(0, "Description");
+        }
+        return placeHoldersList;
+    }
+
+    @Override
+    public List<Row> rows() {
+        List<Row> rows = Lists.newArrayList();
```
l-1squared, line 30: Could the Stream API be used here?

jgiven-core/src/main/java/com/tngtech/jgiven/report/asciidoc/CasesTableImpl.java
l-1squared, line 1: personal note: The contents fo this file come from the `AbstractReportModelHandler`

jgiven-core/src/main/java/com/tngtech/jgiven/report/CasesTable.java
l-1squared, line 1: I am a bit uncertain about this and the `ReportBlockConverter` interface. Right now they are only used for the AsciiDoc generator and I have not yet figured out if they could be used for the other report outlets.

l-1squared (commented):
Unfortunately I did not manage to finish reviewing your PR. There is a lot to look at and a lot to think about here. I have left some initial thoughts though.

jgiven-core/src/main/java/com/tngtech/jgiven/report/CasesTable.java
johthor: Regarding the `CasesTable` as you have already discovered it was part of the remove `AbstractReportModelHandler`. We found it to provide a simple interface for collecting multiple cases into a table. This greatly simplifies reporting on those cases and should be independent from the reporting format used. 

But until now we did not really improve on the implementation of the `CasesTableImpl`. It might be a better approach to collect the table information via a visitor like it is done in the `StatisticsCalculator`. The collection of data should be independent from any reporting format.

The `ReportBlockConverter` on the other hand is part of our reinterpretation of the `AbstractReportModelHandler` class and the `ReportModelHandler` which seemed to provide a format agnostic way to report.  Our idea was for the `ReportBlockConverter` to provide a clear distinction between traversing the model and transforming the information into a specific format. We imagined this to be a starting point for all markup languages where content can be understood as a set of text blocks divided by a block divider like new lines or empty lines. The code might not have communicated the idea in its fullest but a Markdown or reStructured Text reporters could follow the same pattern.

jgiven-core/src/test/java/com/tngtech/jgiven/report/asciidoc/AsciiDocReportBlockConverterTest.java
```diff
@@ -0,0 +1,647 @@
...
+
+        // assert
+        assertThatBlockContainsLines(block,
+            "// tag::scenario-successful[]",
+            "",
+            "==== My first scenario",
+            "",
+            "icon:check-square[role=green] (3s 0ms)",
+            "",
+            "+++Best scenario ever!!!+++",
```
l-1squared, line 162: what kind of formatting do these plusses do?

jgiven-core/src/test/java/com/tngtech/jgiven/report/asciidoc/AsciiDocReportBlockConverterTest.java
```diff
@@ -0,0 +1,647 @@
...
+
+        // assert
+        assertThatBlockContainsLines(block,
+            "===== Case 1 First case",
+            "",
+            "icon:check-square[role=green]");
+    }
+
+    @Test
+    public void convert_case_header_with_one_parameter() {
```
l-1squared, line 202: There doesn't seem to be a diffference between this and the following test.

l-1squared: Also I fail to find the parameters in this test

jgiven-core/src/test/java/com/tngtech/jgiven/report/asciidoc/AsciiDocReportBlockConverterTest.java
```diff
@@ -0,0 +1,647 @@
...
+            "icon:check-square[role=green]");
+    }
+
+    @Test
+    public void convert_step_without_description() {
+        // arrange
+        List<Word> words = ImmutableList.of(Word.introWord("given"), new Word("a coffee machine"));
+
+        // act
+        String block = converter.convertStepBlock(0, words, StepStatus.PASSED, 3899, null, false);
```
l-1squared, line 237: I would put the number `3899` in a global constant name `DURATION` (or DURATION_IN_NANOS). The number doesn't actually matter, but a reader would have an easier time determining its significance

jgiven-core/src/main/java/com/tngtech/jgiven/report/asciidoc/AsciiDocReportBlockConverter.java
```diff
@@ -0,0 +1,432 @@
...
+import com.tngtech.jgiven.report.model.StepStatus;
+import com.tngtech.jgiven.report.model.Word;
+import java.time.Duration;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+
+class AsciiDocReportBlockConverter implements ReportBlockConverter {
+
+    private static final String NEW_LINE = System.getProperty("line.separator");
```
l-1squared, line 21: there's `System.lineSeparator()` which would be even more direct.

jgiven-core/src/main/java/com/tngtech/jgiven/report/asciidoc/AsciiDocReportBlockConverter.java
```diff
@@ -0,0 +1,432 @@
...
+            return stepStatus;
+        }
+    }
+
+    private static String escapeTableValue(final String value) {
+        return escapeArgumentValue(value.replace("|", "\\|"));
+    }
+
+    private static String escapeArgumentValue(final String value) {
+        // TODO Is this really necessary?
```
l-1squared, line 359: that would be great to find out :)

jgiven-core/src/main/java/com/tngtech/jgiven/report/asciidoc/AsciiDocReportBlockConverter.java
```diff
@@ -0,0 +1,432 @@
...
+    private static String generateTableColSpec(final boolean withVerticalHeader, final int columnCount) {
+        return withVerticalHeader
+            ? "h," + generate(() -> "1").limit(columnCount - 1L).collect(joining(","))
+            : generate(() -> "1").limit(columnCount).collect(joining(","));
+    }
+
+    private static String toAsciiDocTag(final ExecutionStatus executionStatus) {
+        switch (executionStatus) {
+            case SCENARIO_PENDING:
+                // fall through
```
l-1squared, line 377: since that is a standard Java feature, I would omit this comment.

jgiven-core/src/main/java/com/tngtech/jgiven/report/asciidoc/AsciiDocReportBlockConverter.java
```diff
@@ -0,0 +1,432 @@
...
+        return generate(() -> symbol).limit(depth + 1L).collect(joining());
+    }
+
+    private static String generateTableColSpec(final boolean withVerticalHeader, final int columnCount) {
+        return withVerticalHeader
+            ? "h," + generate(() -> "1").limit(columnCount - 1L).collect(joining(","))
+            : generate(() -> "1").limit(columnCount).collect(joining(","));
+    }
+
+    private static String toAsciiDocTag(final ExecutionStatus executionStatus) {
```
l-1squared, line 374: How about putting these enum conversion methods in a package-private `EnumValuesMappingUtil` so that this class does not need to concern itsself with the mapping details?

jgiven-core/src/main/java/com/tngtech/jgiven/report/asciidoc/AsciiDocReportBlockConverter.java
```diff
@@ -0,0 +1,432 @@
...
+            case SKIPPED:
+                return ICON_STEP_FORWARD;
+            case PENDING:
+                return ICON_BANNED;
+            default:
+                return stepStatus.toString();
+        }
+    }
+
+    private static Optional<String> toHumanReadableDuration(final long nanos) {
```
l-1squared, line 419: I take issue with this method, because it doesn't exactly announce what it does. (and one may also argue that it does two things), because the method name, in conjunction with the return type suggests that the method will attempt a conversion, which may fail, when in fact it makes a decision whether it should convert AND converts the input.

Unfortunately, I don't have a nifty alternative at hand right now :(

jgiven-tests/src/test/java/com/tngtech/jgiven/report/asciidoc/AsciiDocReportGeneratorTest.java
```diff
@@ -1,50 +1,54 @@
...
-                .and().the_report_exist_as_JSON_file();
+            .and().the_report_exist_as_JSON_file();
 
         when().the_asciidoc_reporter_is_executed();
-        then().the_asciidoc_reporte_$_exists("Test.asciidoc")
-                .and().the_literal_block_is_added_$(
-                "...." + System.lineSeparator() +
-                        content + System.lineSeparator() +
-                        "....");
+        then().the_asciidoc_reporte_$_exists("features/Test.asciidoc")
```
l-1squared, line 47: I kindly want to ask you to fix this old spelling mistake

jgiven-core/src/main/java/com/tngtech/jgiven/report/asciidoc/AsciiDocReportGenerator.java
```diff
@@ -1,196 +1,186 @@
...
+            .sorted(byFeatureName)
+            .forEach(reportModelFile -> writeFeatureToFile(reportModelFile.model, reportModelFile.file,
+                completeReportModel.getStatistics(reportModelFile)));
 
-        @Override
-        public void reportDescription( String description ) {
-            writer.println( description );
-            writer.println();
-        }
+        try (PrintWriter printWriter = PrintWriterUtil.getPrintWriter(new File(targetDir, "allScenarios.asciidoc"))) {
```
l-1squared, line 73: could these `try` statements be split into several methods, like `writeIndex`, `writeGlobalStatistics` etc.?

jgiven-core/src/main/java/com/tngtech/jgiven/report/asciidoc/AsciiDocReportGenerator.java
```diff
@@ -1,196 +1,186 @@
...
+    private File featuresDir;
+    private final Comparator<ReportModelFile> byFeatureName = Comparator.comparing(
+        modelFile -> null != modelFile.model.getName() ? modelFile.model.getName() : modelFile.model.getClassName());
 
-    public void generate() {
-        for( ReportModelFile reportModelFile : completeReportModel.getAllReportModels() ) {
-            writeReportModelToFile( reportModelFile.model, reportModelFile.file );
-        }
-        generateIndexFile();
+    public AsciiDocReportGenerator() {
```
l-1squared, line 46: what is the significance of putting this into the constructor, vs just assigning it at field declaration?

l-1squared (commented):
This is a complex PR :) 
But at least I think I managed to have glance at all the files by now.

l-1squared (commented):
This is a complex PR :) 
But at least I think I managed to have glance at all the files by now.

johthor (commented):
@l-1squared I went through all your comments and incorporated all your suggestions.

Some of them I couldn't believe I missed when creating the PR :grin:

johthor:
I've now reworked the `CasesTable` interface into a simple class without a `CasesTableImpl` and moved the calculation into `CasesTableCalculator`. I'm not to sure about the package `report.model` but every other sub-package seemed equally wrong.

jgiven-core/src/main/java/com/tngtech/jgiven/report/asciidoc/AsciiDocReportGenerator.java
```diff
@@ -1,196 +1,213 @@
...
     }
 
-    class AsciiDocReportModelVisitor implements ReportModelHandler {
-
-        private final PrintWriter writer;
-
-        AsciiDocReportModelVisitor( PrintWriter printWriter ) {
-            this.writer = printWriter;
+    private boolean prepareDirectories() {
+        targetDir = config.getTargetDir();
```
l-1squared, line 72: do we have a test that`targetDir` won't be `null` before it is used in other methods?

jgiven-core/src/main/java/com/tngtech/jgiven/report/asciidoc/AsciiDocReportGenerator.java
```diff
@@ -1,196 +1,213 @@
...
+                writer.println(includeMacroFor(fileName, ""));
+                writer.println();
+            }
         }
+    }
 
-        @Override
-        public void stepEnd() {
-            writer.println( "+" );
+    private void writeFailedScenariosFiled() {
```
l-1squared, line 127: I presume the last word is `Files`

jgiven-core/src/test/java/com/tngtech/jgiven/report/asciidoc/AsciiDocReportModelVisitorTest.java
```diff
@@ -0,0 +1,162 @@
...
+        currentScenarioModel.addCase(new ScenarioCaseModel());
+
+        ReportModel reportModel = new ReportModel();
+        reportModel.addScenarioModel(currentScenarioModel);
+
+        // when
+        reportModel.accept(reportModelVisitor);
+
+        // then
+        Assertions.assertThat(reportModelVisitor.getResult())
```
l-1squared, line 52: could be a static import

l-1squared (commented):
Only  managed to have a short glance today (mostly verified that we can run the tests again). But it looks mostly fine. Maybe we're starting to converge now.

jgiven-core/src/main/java/com/tngtech/jgiven/report/asciidoc/AsciiDocReportGenerator.java
johthor: No there was no test until now and I'm not sure there is even now. I've added some tests but to be honest the different usage patterns with `generate()`, `setConfig` and `loadReportModel` being public and sometimes called from other classes did confuse me a little bit.

I would have preferred for those to not be public.

I hope that `targetDir` being `null` is now handled properly.

jgiven-core/src/main/java/com/tngtech/jgiven/report/model/CasesTable.java
```diff
@@ -0,0 +1,139 @@
...
+     * The rows of the table, not including the header.
+     */
+    public List<CaseRow> rows() {
+        return Collections.unmodifiableList(rows);
+    }
+
+    /**
+     * Represents one case of a scenario.
+     */
+    public static class CaseRow {
```
l-1squared, line 49: yet another case, where a newer java version would have been o so handy.

jgiven-core/src/main/java/com/tngtech/jgiven/report/model/CasesTable.java
```diff
@@ -0,0 +1,139 @@
...
+         * The execution status of the case.
+         */
+        public ExecutionStatus status() {
+            return executionStatus;
+        }
+
+        /**
+         * The duration of the case.
+         */
+        public long getDurationInNanos() {
```
l-1squared, line 118: since all other getters are not prefixed, maybe this one should not be either.

jgiven-core/src/main/java/com/tngtech/jgiven/report/model/CasesTableCalculator.java
l-1squared, line 1: I like this class, it makes the report model visitor look so neat :)

l-1squared (commented):
I managed to get through the productive code today and I like it. 

If I am not mistaken, however, then we need to have a go at the gradle plugin, to make the AsciiDoc report available to gradle users as well

jgiven-core/src/test/java/com/tngtech/jgiven/report/asciidoc/AsciiDocReportConfigTest.java
```diff
@@ -0,0 +1,37 @@
+package com.tngtech.jgiven.report.asciidoc;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import org.junit.Test;
+
+public class AsciiDocReportConfigTest {
```
l-1squared, line 7: How about another test with an input "--sourceDir=". We currently fail gracefully and print the help message, which I think is good behavior.

jgiven-core/src/test/java/com/tngtech/jgiven/report/asciidoc/AsciiDocReportGeneratorTest.java
```diff
@@ -0,0 +1,76 @@
+package com.tngtech.jgiven.report.asciidoc;
```
l-1squared, line 1: Shouldn't we at least have some test here that checks that we have actually generated a report?

jgiven-core/src/test/java/com/tngtech/jgiven/report/asciidoc/AsciiDocReportModelVisitorTest.java
```diff
@@ -0,0 +1,163 @@
...
+                        "convertScenarioHeaderBlock",
+                        "convertCaseHeaderBlock",
+                        "convertFirstStepBlock",
+                        "convertStepBlock",
+                        "convertFirstStepBlock",
+                        "convertCaseHeaderBlock",
+                        "convertScenarioFooterBlock"));
+    }
+
+    private static class MyFakeReportBlockConverter implements ReportBlockConverter {
```
l-1squared, line 110: I am curious why you chose this implementation over just using mockito. Is there a specific reason?

jgiven-core/src/test/java/com/tngtech/jgiven/report/asciidoc/AsciiDocReportModelVisitorTest.java
```diff
@@ -0,0 +1,163 @@
...
+                        "convertFeatureHeaderBlock",
+                        "convertScenarioHeaderBlock",
+                        "convertCaseHeaderBlock",
+                        "convertFirstStepBlock",
+                        "convertCaseHeaderBlock",
+                        "convertScenarioFooterBlock"));
+    }
+
+    @Test
+    public void visitTheReportModelWithTwoSections() {
```
l-1squared, line 64: shouldn't this be spelt `visitAScenarioWithTwoScenarioCases`? Speaking of which, checking that two Scenarios in a report are handled gracefully might also be a good idea

l-1squared (commented):
More review. I'm halfway through the test files now :)

jgiven-core/src/test/java/com/tngtech/jgiven/report/asciidoc/AsciiDocStepBlockConverterTest.java
```diff
@@ -0,0 +1,218 @@
...
+        // then
+        assertThatBlockContainsLines(block,
+            "* [.jg-intro-word]*Given* a coffee machine +",
+            "  _+++It is a brand new machine.+++_");
+    }
+
+    @Test
+    @DataProvider({"PASSED, check-square[role=green]", "FAILED, exclamation-circle[role=red]",
+        "SKIPPED, step-forward[role=silver]", "PENDING, ban[role=silver]"})
+    public void convert_step_from_unsuccessful_scenario_case(final StepStatus stepStatus, final String icon) {
```
l-1squared, line 55: Given that the data provider also includes "passed", "unsuccesful" is inapproproate here. How about "converts_scenario_cases_step_status"?

jgiven-core/src/test/java/com/tngtech/jgiven/report/model/CasesTableCalculatorTest.java
```diff
@@ -0,0 +1,173 @@
...
+
+        // when
+        final CasesTable casesTable = calculator.collectCases(scenario);
+
+        // then
+        assertThat(casesTable.hasDescriptions()).isTrue();
+        assertThat(casesTable.placeholders()).containsExactly("foo", "bar");
+        assertThat(casesTable.rows())
+                .hasSize(2)
+                .satisfiesExactly(
```
l-1squared, line 96: Oh! I didn't know that feature.

l-1squared (commented):
@johthor  I made it through! 
I'm happy about this PR.

jgiven-gradle-plugin/src/main/java/com/tngtech/jgiven/gradle/internal/AbstractJGivenReportImpl.java
```diff
@@ -57,7 +57,7 @@ public AbstractReportGenerator createGenerator() {
         if( getTitle() != null ) {
             conf.setTitle( getTitle() );
         }
-        conf.setTargetDir( getDestination() );
+        conf.setTargetDir( getOutputLocation().getAsFile().get() );
```
l-1squared, line 60: another gradle deprecation?

l-1squared (commented):
Looks fine at a glance. Will do a final sweep on monday

jgiven-core/src/main/java/com/tngtech/jgiven/report/asciidoc/AsciiDocReportGenerator.java
```diff
@@ -1,196 +1,226 @@
...
+            failedScenarioFiles.add(featureFileName);
+        }
+        if (statistics.numPendingScenarios > 0) {
+            pendingScenarioFiles.add(featureFileName);
+        }
 
-            for( ScenarioDataTable.Row row : scenarioDataTable.rows() ) {
-                writer.print( "| " + row.nr() );
+        final AsciiDocReportModelVisitor visitor = new AsciiDocReportModelVisitor(blockConverter, statistics);
+        reportModelFile.model.accept(visitor);
```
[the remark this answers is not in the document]
johthor: I cant't answer the question for that code from 2014. Until now I tried not to change anything I did not write myself or absolutely had to modify.

But I will gladly change that class if you want me to.

jgiven-core/src/test/java/com/tngtech/jgiven/report/asciidoc/AsciiDocReportConfigTest.java
johthor: The `ConfigOptionParser` calls `System.exit()` in this case and currently I'm not sure how to test for that case without introducing a lot of additional complexity.

jgiven-core/src/test/java/com/tngtech/jgiven/report/asciidoc/AsciiDocReportGeneratorTest.java
johthor: There is the identical named `jgiven-tests/src/test/java/com/tngtech/jgiven/report/asciidoc/AsciiDocReportGeneratorTest.java` which checks that the five meta files and one feature file is generated. Which was enough for me. Or to be honest I didn't want to handle files and file contents in the tests.

jgiven-core/src/test/java/com/tngtech/jgiven/report/asciidoc/AsciiDocReportModelVisitorTest.java
johthor: Mockito was not part of the test dependencies of the `jgiven-core` project and I didn't want to add it if it wasn't already.
I presumed there was a reason for it to not be added. So I implemented my own _order of invocation verifier_.

jgiven-core/src/test/java/com/tngtech/jgiven/report/asciidoc/AsciiDocScenarioBlockConverterTest.java
```diff
@@ -0,0 +1,122 @@
...
+        "SCENARIO_PENDING, pending, icon:ban[role=silver]",
+        "SOME_STEPS_PENDING, pending, icon:ban[role=silver]"})
+    public void convert_scenario_header_without_tags_or_description(final ExecutionStatus status,
+                                                                    final String scenarioTag,
+                                                                    final String humanStatus) {
+        // given
+        List<String> tagNames = new ArrayList<>();
+
+        // when
+        String block = converter.convertScenarioHeaderBlock("my first scenario", status, 1000300000L, tagNames, null);
```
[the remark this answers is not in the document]
johthor: Those are 300000 nanoseconds so there are no milliseconds to be displayed. I've now introduced underscores to better communicate the different time units and extracted variables.

jgiven-core/src/test/java/com/tngtech/jgiven/report/asciidoc/AsciiDocStepBlockConverterTest.java
johthor: No, the unsuccessful does not described the status of the steps execution but the execution status of the scenario case it is included in.
In successful scenarios and scenario cases there is no need to print every step's execution status as they should all be `passed` only for non successful scenario cases the steps status should be reported.

johthor: But that comment showed me that the status reporting was not implemented correctly for all parameter types ;(

jgiven-core/src/test/java/com/tngtech/jgiven/report/asciidoc/AsciiDocReportModelVisitorTest.java
```diff
@@ -0,0 +1,296 @@
...
+                        "StepBlock",
+                        "ScenarioFooterBlock"));
+    }
+
+    @Test
+    public void visits_a_report_with_two_scenarios() {
+        // given
+
+        ReportModel report = mkReport(
+                mkScenario("Scenario One", false, mkScenarioCase(
```
l-1squared, line 59: Its really rare to see helper /shortcut methods in a test that work that well. Nice :)

l-1squared (commented):
Hi, 
so codewise I think we're fine. 
But I noticed that there are failing test (for me locally) in core and jgiven-tests. 
The one in core is somewhat my fault as I asked you to write one with bogus input for the AsciiDocConigBuilder (or a similar name). I guess we need to scrap that. 

then in test, my candidate for most annoying test in the codebase seems to fail for you, which seems to stem from the fact that you are not using the current master as base. 

Finally, I took a look and the ascii-report generated for JGiven tests and I noticed two things.
* in com.tngtech.jgiven.report.analysis.ArgumentAnalyzerTest.asciidoc the tag of the first test doesn't get processed properly and loses its number. I'd say that this should be fixed, before this awesome is released
* Secondly I think that the hints as such (i.e. text written in cursive, see same test), may be only recognizeable to the initiated. I was wandering wheter placing a "hint" or "meaning" before that hint and/or putting the hint in parentheses will help here.

jgiven-core/src/main/java/com/tngtech/jgiven/report/asciidoc/AsciiDocReportGenerator.java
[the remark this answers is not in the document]
l-1squared: Seems you did, and that was a good little refactoring.

jgiven-gradle-plugin/src/main/java/com/tngtech/jgiven/gradle/internal/AbstractJGivenReportImpl.java
johthor: yes.

jgiven-core/src/main/java/com/tngtech/jgiven/report/asciidoc/AsciiDocSnippetGenerator.java
```diff
@@ -0,0 +1,62 @@
...
+
+    private List<String> generateIndexSnippet(final String intro, final String tags) {
+        final ArrayList<String> result = new ArrayList<>();
+        result.add(intro);
+
+        if (!Strings.isNullOrEmpty(tags)) {
+            result.add(":leveloffset: -1");
+        }
+
+        for (final String fileName : featureFiles) {
```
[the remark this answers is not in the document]
johthor: Yeah sure.

johthor:
I must already have removed the failing `AsciiDocReportConfigTest` again after discovering that `System.exit()` was involved :grin: .

And I now have merged the current master. I hope that will take care of some of those failures.

I've introduced a :heavy_plus_sign: in front of the extended step descriptions to hopefully communicate the meaning better. I didn't like the idea of introducing text for explanation or parentheses. The icon also got a title that should display on hover. Unfortunately there is no tool tip in AsciiDoc like in the HTML report.

And I will try fix the handling of tags. I did know that the implementation is lacking there.

jgiven-core/src/test/java/com/tngtech/jgiven/report/asciidoc/AsciiDocReportModelVisitorTest.java
johthor: Yeah the test did check multiple things. Now there are more tests that hopefully only check for exactly one situation.

johthor:
For me locally only tests from `Html5AppTest` from `jgiven-tests` are failing and I don't fully understand why they are failing.

jgiven-gradle-plugin/src/main/java/com/tngtech/jgiven/gradle/internal/AbstractJGivenReportImpl.java
l-1squared: thanks for resolving

l-1squared:
> I must already have removed the failing `AsciiDocReportConfigTest` again after discovering that `System.exit()` was involved 😁 .
> 
> And I now have merged the current master. I hope that will take care of some of those failures.
> 
> I've introduced a ➕ in front of the extended step descriptions to hopefully communicate the meaning better. I didn't like the idea of introducing text for explanation or parentheses. The icon also got a title that should display on hover. Unfortunately there is no tool tip in AsciiDoc like in the HTML report.
> 
> Handling tags with values and without values was less complex than I thought. And now there is also support for `cssClass`.

As I said, the test is a total nuissance. Yesterday (bevore merging) it failed on my setup because it was still hardwired to the chrominum driver, which currently seems to have a bug that prevents chromium (and maybe all chromium based browsers) to access any websit if not some "origins allowed is everything" header is set. 
I (and I think the CI) don't see these problems, because the test first tries out which browsers are present in the order firefox -> chrome -> edge. (Reduces reproducability, but then again the choice of browser should not matter for the test proper and what information is there to be had when you fail, because the desired browser was not available). So maybe you fail because the test's browser selecter ends up using chrome whereas the CI and I end up using firefox.

l-1squared (dismissed):
I see that you have not yet completed your list of features that you'd like to have. 
But maybe it makes sense to add these things as a new PR

johthor:
YES, that would be the better choice.

Printing the parameters for standalone scenarios was done in the original implementation, to behave like the HTML report. But I decided to remove it as the standalone cases do print the parameters values inline anyways.

The duration for scenario cases I thought would not be shown in the HTML report which seems to be dependent on the situation. I think I want to still add it, it shouldn't be a big change.

The Attachments I won't add in this PR, as I'm not sure if it is worth to re-implement the `HTML5AttachmentGenerator` which is unfortunately only part of the `jgiven-html5-report` project.

johthor:
And I'm done. :grin: 

I've updated the PR description accordingly and even improved the rendering of tables a little bit.
I don't want to get any more _quick_ ideas in this PR.

l-1squared (commented):
Hi,  

I have noticed some issues while running the test on my local machine:
* There are some failing tests in jgiven-core on my machine:
** AsciiDocScenarioCaseBlockConverterTest#convert_scenario_case_footer_without_stacktrace
** AsciiDocTableBlockConverterTEst#conert_cases_table_with_errors
** AsciiDocStepBlockConverterTest

This seems to be an issue with line breaks which  for some reason I can see Unix line breaks  on my Windows machine, but I have also noticed, that at least some tests just have far too few of them.

Another, more troublesome matter is that I did not find a duration in the asciidocReport, where I expected it. Case  in point: jgiven-tests, HTML App->Newlines are detected in formatted values and shown as multiline text, Step "When the HTML Report Generator is executed" and the following one. They do have durations on them in the HTML App, but not in the ascii doc report. 
Do you see the same behavior?

johthor:
Hi @l-1squared,

unfortunately I completely forgot about the line ending problem with Windows and Linux/MacOS and did not use the `System.lineSeparator()` in all places. That should hopefully be fixed now. I'm missing a Windows development machine so I can't be sure. I did use `\n` in test inputs but the code checked for `System.lineSeparator()` which was `\r\n` on your machine.

Other parts of the code solved this problem by using `actual.replaceAll(System.getProperty("line.separator"), "\n")`. I'm not sure which way is simpler to understand.

Regarding the step duration I could reproduce the outcome => I did misunderstand the display of durations in the HTML report again. ;(

I've also reworked that part. Now step status and step duration may be displayed independent from one another. Step status is only displayed if the surrounding scenario (case) was unsuccessful. Step duration is only displayed if it is above 10,000,000 nanoseconds or 10 milliseconds.

The status and duration of scenarios and scenario cases is always displayed.

l-1squared (approved):
@johthor  I'll let you do the honors

(do you actually have permission to do so?)

johthor:
I would love to click the BIG BUTTON of happiness but I may only _close with comment_ or _comment_ on this PR.

So you will have to merge the PR as I don't have write permissions.

l-1squared:
> @l-1squared I would love to click the BIG BUTTON of happiness but I may only _close with comment_ or _comment_ on this PR.
> 
> So you will have to merge the PR as I don't have write permissions.

Done! (Took a while, I was on a business trip)

PR #1353, opened 2023-07-24
Add initial documentation for the AsciiDoc reporter

johthor:
Describe how the AsciiDoc report may be generated via CLI or Maven.

Unfortunately I couldn't describe the Gradle configuration.

The sections for the Maven and Gradle Plugin will be upgraded to level 2 because it didn't make sense to describe them twice. Once for HTML and once for Asciidoc.

Issue: #54


l-1squared (approved):
Thanks a lot for cleaning up this file

l-1squared:
(!) @johthor  can this be merged? 
I'd like to do a release (we have a bug open that was already fixed on master) and frankly we should have done one already. But I don't want to cut out your contribution from the release (would make a real mess of the commit graph). Rather I think it makes more sense to add this docs with a mark  "Experimental"

l-1squared:
Hi @johthor   
I do know that life can easily get in the way of open source contributions. I want to ask, whether you still have this project on the radar. It would be a pity if your significant contribution remained incomplete.  It also pains me that I feel I cannot honor your efforts in the Changelog, because that would mean officialy releasing the feature, which I can't because it is incomplete.

PR #1423, opened 2023-10-18
Issue-1420: Update jansi to version 2.4x

l-1squared:
I made some simple assumptions about the nature of the parameters of the output stream. Notably -> REST_CODE is RESET_CODE

### DIFF
--- a/jgiven-core/src/main/java/com/tngtech/jgiven/impl/util/AnsiUtil.java
+++ b/jgiven-core/src/main/java/com/tngtech/jgiven/impl/util/AnsiUtil.java
@@ -1,13 +1,13 @@
 /**
  * Copyright (C) 2009, Progress Software Corporation and/or its
  * subsidiaries or affiliates. All rights reserved.
- *
+ * <p>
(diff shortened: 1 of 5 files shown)


jgiven-core/src/main/java/com/tngtech/jgiven/impl/util/AnsiUtil.java
```diff
@@ -1,13 +1,13 @@
 /**
  * Copyright (C) 2009, Progress Software Corporation and/or its
  * subsidiaries or affiliates. All rights reserved.
- *
+ * <p>
```
fudler, line 4: Uh, what's the point of this? :)

fudler (approved):

jgiven-core/src/main/java/com/tngtech/jgiven/impl/util/AnsiUtil.java
l-1squared: checkstyle likes it better that way, I guess...

PR #1424, opened 2023-10-18
Use java 21 on test systems

l-1squared:


fudler (approved):

l-1squared:
Does not work right now bytebudd and android? don't support java21 yet

PR #1460, opened 2023-11-24
Issue-802: Test maven mojo manually

l-1squared:


jgiven-maven-plugin/src/test/java/com/tngtech/jgiven/maven/BasicMavenMojoTest.java
```diff
@@ -1,85 +1,119 @@
 package com.tngtech.jgiven.maven;
 
+import org.apache.maven.plugin.MojoExecutionException;
 import org.apache.maven.shared.invoker.*;
```
fudler, line 4: It just comes to my mind, do we need the *-import here?

jgiven-maven-plugin/src/test/java/com/tngtech/jgiven/maven/BasicMavenMojoTest.java
```diff
@@ -1,85 +1,119 @@
...
 
-    @Test
-    void testMavenProducesHtmlAndAsciiReport() throws MavenInvocationException {
-        var mavenExecutable = this.findMavenExecutable();
+    @ParameterizedTest
+    @MethodSource("provideFormatsAndExpectedOutput")
+    void should_determine_output_dir_based_on_format(String format, String expectedOutput) throws MojoExecutionException {
+        var mavenMojo = new JGivenReportMojo(
+                new File(projectFile, "target"),
+                null,
```
fudler, line 52: I suppose we can't avoid a null here?

fudler (approved):
Approved but left two comments.

jgiven-maven-plugin/src/test/java/com/tngtech/jgiven/maven/BasicMavenMojoTest.java
l-1squared: In fact, that particular `null` is even necessary, because we're testing that the MOJO chooses the correct output directory

PR #1542, opened 2024-02-19
Feature/separate plugin config and execution

l-1squared:
TODO:
* test that the jgiven report doesn't produce results if neither test, nor report is requested
* changelog
* credit in changelog


jgiven-gradle-plugin/src/test/java/com/tngtech/jgiven/gradle/JGivenPluginTest.java
l-1squared, line 1: reconsider what the formatter did here :(

jgiven-gradle-plugin/src/test/java/com/tngtech/jgiven/gradle/JGivenPluginTest.java
```diff
@@ -62,111 +64,112 @@ public void plugin_can_build_with_an_empty_project() throws Exception {
...
-            .the_plugin_is_applied()
-            .and().there_are_JGiven_tests();
+                .the_plugin_is_applied()
+                .and().there_are_JGiven_tests();
 
         when()
-            .a_build().with()
-            .the_task("jgivenTestReport").is_successful();
+                .a_build().with()
+                .the_task("-x").the_task("test").and()
```
l-1squared, line 168: @jjohannes I'd like to, for one final time, have your opinon on this one. 
To me it seems, now that we properly use Providers, Gradle is able to track the  necessary inputs of the JGiven task. 
Therefore, previously a JGiven report task would not run, if a test task was not explicitely requested. Now, I assume that because we explicitely request the report task to run, Gradle knows that it needs to run  the test task and does so accordingly.

While this is changed behavior, I am of the opinon, that we are now more in line with how Gradle is expected to behave and therefore this would generally be an appreciated change.

How do you see this?

jjohannes: Yes, your observation is correct. The providers that point at task outputs carry the task dependency information and that is why the `test` task is automatically executed before.

If you can accept the breaking change, I would agree that this is the better behavior. It follows Gradle best practices. (If not, there would be ways to "trick" the setup into loosing the dependency but still using providers in general.)

Note: Traditionally the `jacoco` report task built into Gradle core also does not follow this best practice. However I believe this is due to not wanting to break the behavior that was introduced early (pre Gradle 1.0). If the JaCoCo tasks/plugin would be created today, I think it would also behave different.

jgiven-gradle-plugin/src/main/java/com/tngtech/jgiven/gradle/JGivenPlugin.java
```diff
@@ -79,20 +78,18 @@ private void addDefaultReports(final Project project) {
...
-            });
+
+        Provider<Directory> getResultsDirectory = test.getExtensions()
+                .getByType(JGivenTaskExtension.class)
+                .getResultsDir();
+
+        reportTask.getResults().set(getResultsDirectory);
+
+        String relativeFilePath = "jgiven" + "/" + test.getName() + "/";
+        Provider<Directory> reportOutputLocation= reportingExtension.getBaseDirectory().dir(relativeFilePath);
```
jjohannes, line 89:
```suggestion
        Provider<Directory> reportOutputLocation = reportingExtension.getBaseDirectory().dir(relativeFilePath);
```

jjohannes (approved):

jgiven-gradle-plugin/src/main/java/com/tngtech/jgiven/gradle/JGivenReportTask.java
```diff
@@ -20,29 +22,22 @@
...
     public JGivenReportTask() {
-        reports = getInstantiator().newInstance(JGivenReportsContainerImpl.class, this);
+        reports = getObjects().newInstance(JGivenReportsContainerImpl.class, this);
     }
 
     @Inject
-    protected Instantiator getInstantiator() {
+    protected ObjectFactory getObjects() {
         throw new UnsupportedOperationException();
     }
```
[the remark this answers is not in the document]
jjohannes: Just a suggestion to make it "more compact".

jgiven-gradle-plugin/src/main/java/com/tngtech/jgiven/gradle/JGivenPlugin.java
```diff
@@ -79,20 +78,18 @@ private void addDefaultReports(final Project project) {
...
-                String relativeFilePath = "jgiven" + "/" + test.getName() + "/" + report.getName();
-                Callable<File> getDestination = () -> reportingExtension.file(relativeFilePath);
-                reportMapping.map("destination", getDestination);
-            });
+
+        Provider<Directory> getResultsDirectory = test.getExtensions()
+                .getByType(JGivenTaskExtension.class)
+                .getResultsDir();
+
+        reportTask.getResults().set(getResultsDirectory);
```
jjohannes, line 86:
```suggestion
        Provider<Directory> resultsDirectory= test.getExtensions()
                .getByType(JGivenTaskExtension.class)
                .getResultsDir();

        reportTask.getResults().convention(resultsDirectory);
```

Using `convention` makes this the "fall-back" value. Then you can be sure that if a user re-configures this it will override the default. Changing the value here also allows users to "break" the task dependency to the test task if they are not happy with that change.

l-1squared: Funnily enought, I had this as convention for a while, but convinced myself that "set" was correct ^^

jgiven-gradle-plugin/src/test/groovy/com/tngtech.jgiven.gradle/JGivenPluginShould.groovy
```diff
@@ -18,6 +18,37 @@ class JGivenPluginShould extends Specification {
...
+    """
+
+        new File(testProjectDir,"src/test/java").mkdirs()
+        File scenario = new File(testProjectDir,"src/test/java/SimpleScenario.java")
+        scenario << Resources.toByteArray(Resources.getResource("SimpleScenario.java"))
+
+        when:
+        def result = GradleRunner.create()
+                .withProjectDir(testProjectDir)
+                .withArguments("--configuration-cache", "test", "jgivenTestReport", "-S", "--info"/*, "-Dorg.gradle.debug=true"*/ )
```
fudler, line 43: Will this comment still have some purpose?

jgiven-gradle-plugin/src/main/java/com/tngtech/jgiven/gradle/JGivenPlugin.java
```diff
@@ -3,23 +3,25 @@
...
-import java.util.Objects;
-import java.util.concurrent.Callable;
-import org.gradle.api.Action;
-import org.gradle.api.NonNullApi;
-import org.gradle.api.Plugin;
-import org.gradle.api.Project;
-import org.gradle.api.Task;
-import org.gradle.api.internal.ConventionMapping;
-import org.gradle.api.internal.IConventionAware;
+import org.gradle.api.*;
```
fudler, line 6: Could this import be avoidable?

fudler (approved):
Two small comments added.

jgiven-gradle-plugin/src/test/groovy/com/tngtech.jgiven.gradle/JGivenPluginShould.groovy
l-1squared: Damn, I totally missed that.  but @jjohannes  thanks for the suggestion

PR #1719, opened 2024-09-06
JGiven-23: draft documentation for plugin docs

fudler:


RobertSchueler (approved):

_includes/.head.html.swp
l-1squared, line 1: please don't commit vi backup files

l-1squared (changes requested):
What happened here?

_includes/.head.html.swp
fudler: I'm not too sure what went wrong. My initial commit consisted of two files and... well, one commit.

PR #1797, opened 2024-12-03
fix(Workflows): Replace deprecated wrapper validation

l-1squared:
As per recommendation from gradle


l-1squared:
Too trivial for review

PR #1875, opened 2025-02-27
Fix/issue 1856 aborted is not success

l-1squared:


jgiven-junit5/src/main/java/com/tngtech/jgiven/junit5/JGivenExtension.java
```diff
@@ -87,6 +83,8 @@ public void afterTestExecution(ExtensionContext context) throws Exception {
         try {
             if (context.getExecutionException().isPresent() && !ThrowableUtil.isAssumptionException(context.getExecutionException().get())) {
                 scenario.getExecutor().failed(context.getExecutionException().get());
+            } else if (context.getExecutionException().isPresent() && ThrowableUtil.isAssumptionException(context.getExecutionException().get())) {
+                scenario.getExecutor().aborted(context.getExecutionException().get());
```
RobertSchueler, line 87: Probably this would be a bit easier to read with a submethod that uses an early return, i.e.

```
if (!context.getExecutionException().isPresent()) { return;}
if (ThrowableUtil.isAssumptionException(context.getExecutionException().get())) {
  scenario.getExecutor().aborted(context.getExecutionException().get());
  return;
}
scenario.getExecutor().failed(context.getExecutionException().get());
```

RobertSchueler (dismissed):

jgiven-junit5/src/main/java/com/tngtech/jgiven/junit5/JGivenExtension.java
l-1squared: Nice.
To lazy to refactor, immediately caught :D

RobertSchueler (approved):

PR #1997, opened 2025-07-29
Feature/new publish plugin

l-1squared:


.github/workflows/verify_signing.yml
```diff
@@ -0,0 +1,124 @@
...
+on:
+  - push
+  - pull_request
+
+jobs:
+  verify-signing:
+    name: Verify JGiven Signing
+    runs-on: ubuntu-latest
+    env:
+      GPG_PASSWORD: 'test-signing-password'
```
RobertSchueler, line 11: Is it ok to have these as plain text?
Maybe should be Github secrets - or would you say for testing it is ok?

PR #2244, opened 2026-06-23
chore(Core): Avoid injection errors for Java 26

### LINKED ISSUE #2224: JGiven breaks on Java 26+ with the newest (1.18.10) byte-buddy
The newest BB has this: https://github.com/raphw/byte-buddy/commit/f3148185ceedb74a3b7383b6f706fa99fca4c438 and JGiven broke for me on Java 25 now, once I bumped BB.

Per the [BB comment](https://github.com/raphw/byte-buddy/commit/f3148185ceedb74a3b7383b6f706fa99fca4c438#r187449549):
>  You would need to avoid this strategy and use `UsingLookup` instead which is the only approach that the JVM supports in the near future.
>  

 _Originally posted by @pzygielo in [#370](https://github.com/TNG/JGiven/issues/370#issuecomment-4612635892)_

l-1squared:


jgiven-core/src/test/java/com/tngtech/jgiven/impl/ScenarioExecutorTest.java
hvennekate, line 1: General comment on this class: it would probably be better to repeat the test cases for the different strategies in more general terms: First, I noticed that the tests are essentially the same -- as are the stage classes (it does not seem to me that this is really necessary given that the executor retrieves the actual instances from  the creator anyway). At that point, I would have suggested using nested test classes to group the test cases. But then realizing that they are essentially (did not compare each, tbh.) the same, it seems to me like it should be possible to have one generalized test base and just inherit that in each of those nested classes with some specialized setup for each strategy, no?

hvennekate (approved):

example-projects/java26-module/README.md
```diff
@@ -0,0 +1,55 @@
+# Java 21 Module Example Project
```
hvennekate:
```suggestion
# Java 26 Module Example Project
```

example-projects/java26-module/README.md
```diff
@@ -0,0 +1,55 @@
...
+`--patch-module`s the compiled test classes into that module.
+
+## Run
+
+1. Publish the JGiven artifacts to a local Maven repository, or point
+   `version` in `build.gradle.kts` at a released version.
+2. From this directory:
+
+   ```
+   ../../gradlew build
```
hvennekate, line 38: I did get
```
example-projects/java26-module/src/test/java/module-info.java:19: warning: [module] module not found: net.bytebuddy
```
But it does not build without it, so I suppose it is not picking up the transitive dependency correctly at that stage?

jgiven-core/src/main/java/com/tngtech/jgiven/impl/ByteBuddyStageClassCreator.java
```diff
@@ -1,66 +1,172 @@
...
+ *       package membership. This tier is reached on Java 9+ runtimes that
+ *       have not opened {@code java.base/java.lang} to reflection. If
+ *       {@code privateLookupIn} fails (typically because the user's JPMS
+ *       module does not {@code --add-opens} its package to JGiven), a
+ *       single {@code WARN}-level log message is emitted and the strategy
+ *       falls through to {@code WRAPPER}.</li>
+ *
+ *   <li><strong>Wrapping</strong> ({@link ClassLoadingStrategy.Default#WRAPPER})
+ *       &mdash; fallback when neither injector is available, or when the
+ *       lookup tier cannot acquire a usable {@code Lookup}. The generated
```
hvennekate, line 53: Or when the stage's classloader cannot be determined?

jgiven-core/src/main/java/com/tngtech/jgiven/impl/CachingStageClassCreator.java
```diff
@@ -20,4 +20,8 @@ public <T> Class<? extends T> createStageClass( Class<T> stageClass ){
         return (Class<? extends T>) typeCache.findOrInsert( stageClass.getClassLoader(), stageClass,
                 () -> stageCreator.createStageClass( stageClass ) );
     }
+
+    static void clearCache() {
```
hvennekate, line 26: We don't do `@VisibleForTesting`?

hvennekate (approved):

jgiven-core/src/main/java/com/tngtech/jgiven/impl/ByteBuddyStageClassCreator.java
l-1squared: when would that be the case?

jgiven-core/src/main/java/com/tngtech/jgiven/impl/CachingStageClassCreator.java
l-1squared: Totally forgot that it exists.

example-projects/java26-module/README.md
l-1squared: My guess would be that the error happens because gradles keeps the dep out of the module path. But at runtime bytebuddy of course needs this access

l-1squared: Jup. But I think the correct course of action here ist to do nothing.

